PR #45, opened 2024-04-04
Translation to Spanish

cvc90:
The translation file has been created in the Spanish language and its operation has been validated

image


vingerha:
Since coming from the Emperor (supposedly you died a.d. 1558) a special thanks :)

PR #47, opened 2024-04-04
Update config_flow.py

cvc90:
Currently the variable "CONF_RADIUS" is set with min=50, max=500, I think that for cities it is correct but for small towns or remote houses I think it should be possible to increase the range to 1500M or even more.

Increase maximum range in "CONF_RADIUS" to 1500M


vingerha:
Rejecting the proposal as it is not that simple to raise the limit wihtout a posisble system-killing risk. See next release for an alternative solution

PR #77, opened 2024-05-22
Split code.

FabienD74:
Original version

### DIFF
--- a/custom_components/gtfs2/sensor.py
+++ b/custom_components/gtfs2/sensor.py
@@ -58,6 +58,8 @@ from .const import (
 )
 from .coordinator import GTFSUpdateCoordinator, GTFSLocalStopUpdateCoordinator
 
+from .sensor_stoplist import *
+
 _LOGGER = logging.getLogger(__name__)
 
 
--- a/custom_components/gtfs2/sensor.py
+++ b/custom_components/gtfs2/sensor.py
@@ -1,7 +1,11 @@
 """Support for GTFS."""
-from datetime import datetime
+from datetime import datetime, timezone
+
 import logging
 from typing import Any
+
+import os
+import geopy.distance
 
 from homeassistant.components.sensor import SensorDeviceClass, SensorEntity
 from homeassistant.config_entries import ConfigEntry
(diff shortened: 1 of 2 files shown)


vingerha:
You are changing things ...do let me know when this is clean from device-tracker related things and ready for review, no time to check over and over again :)

vingerha:
Deploying PR is a not as straightforward ... I know.... just let me know when done
Another comment...  the (new) local stop sensor list will eventually pose issues as it will grow out of acceptable size if you are in a big town and/or big radius....

FabienD74:
i have no quick easy solution.
If u think there is a risk, then may be we should revert everyting back... and delete this PR.

FabienD74:
I will continue on my fork.... I still need to find out the "destination" of the route. Otherwise this whole development is useless.
Regards

vingerha:
Yep...do continue and you can run all you build for yourselves of course. I can also have a look at the fork when you think you are done..merging with my main needs a reasonably OK basis (perfect does not exist :) ). For the local list, yes...this is going to be a possible issue...this is the 'warning' that will follow in such cases, not sure if it truncates 
```
.... exceed maximum size of 16384 bytes. This can cause database performance issues; Attributes will not be stored
` `

vingerha:
> I will continue on my fork.... I still need to find out the "destination" of the route. Otherwise this whole development is useless. Regards

I can only say good-luck but I have to admit that I personally donot dee that happening

FabienD74:
Thanks for your research.... => 16Kb ... that should be enough for my needs...
It's already a huge timetable if I remove some info like route_name, coords,...

What a week!!! 
First dev in python 👍, first tme with github 👍, and first "custom component" ! 👍

That's too much in parallel ;-) ;-) ;-)  
Sorry

vingerha:
Further small test.. I added a zone in amsterdam with many stop sbut there is no stop and no list, als the one that showed above 10k before no longer pops-up with data. I will revert this as it is not stable... thorough testing is required first. 
I will add a DEV branch so you can sync with that

image

vingerha:
New branch dev created on the 'main' before applying your previous PR, I can create releases either from main or dev so this at least keeps 'main' as clean as can be ... and donot get me wrong, I do have to return quite often on my code :)
Still, I propose you first create a working version on your fork, test this for zone and person and for multiple sources ....  and when ready I will test it out as well

FabienD74:
Perfect!

vingerha:
btw... to avoid making this a semi-chat post ...I am on disocrd with the same name, easier and less polluting for others :)

vingerha:
Closing as this has mostly been used for chatty stuff and the PR is not complete

PR #91, opened 2024-06-13
Add support for an API Key on the static data source

mark-lawrenson:
Hi there,

This PR adds support for an API Key to be used when accessing the static data source URL, in the same way as it is possible with the real time feeds.

Transport for NSW (Australia) has a seperate set of timetables which correspond with the real time feeds, and these timetables are only accessible with an API Key. With these changes, it is possible to configure the integration using the URL for those timetables, rather than the complete static file that does not correspond with the real time feeds.

I'm not sure if this will be common with any other transport operators, but it is of course still possible to not use an API Key, as the implementation is identical to how it was done for the real time feeds.

Details:
* Introduces API Key configurations to the data schema for the new data source step.
* Adds corresponding translations (using the translation data already available from the real time step)
* Adds support in `get_gtfs` to use the API Key as configured in the GET request

Kind regards,
Mark

### DIFF
--- a/custom_components/gtfs2/config_flow.py
+++ b/custom_components/gtfs2/config_flow.py
@@ -147,13 +147,38 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
                 step_id="source",
                 data_schema=vol.Schema(
                     {
-                        vol.Required(CONF_EXTRACT_FROM): selector.SelectSelector(selector.SelectSelectorConfig(options=["url", "zip"], translation_key="extract_from")),
+                        vol.Required(CONF_EXTRACT_FROM): selector.SelectSelector(
+                            selector.SelectSelectorConfig(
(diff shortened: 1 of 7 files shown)


vingerha:
Thanks for this, it was on my list but the only other static api-needing source till now (that I know of) was Sweden and there you can put the api in the url so...no need = no prio .... that was my take on it. 
I would like to ask you for one other thing and that is to add this to the service as well, the service allows you to update manually of via automation...hence 

image


same / similar as this one 

image

mark-lawrenson:
@vingerha Awesome, thanks for mentioning that, I didn't notice the services! I've added that functionality now.

vingerha:
I added it to 'main' now which will have a few other things that I am testing out too which will take a week or so to validate (I do have another life too). Expecting a new release by/in the weekend ...thanks again

PR #153, opened 2026-03-10
Fix Python 3.14 process pickling crash and SQLAlchemy fd leak

dafal:
## Summary

This PR fixes two related issues that cause instability under Home Assistant 2026.3 (Python 3.14).

### Issue 1 — Process args pickling crash (TypeError: cannot pickle '_asyncio.Future' object)

**Root cause**: Python 3.14 changed the default multiprocessing start method from `fork` to `forkserver` on POSIX ([docs](https://docs.python.org/3.14/library/multiprocessing.html#contexts-and-start-methods)). `forkserver` requires all `Process` args to be picklable, but `hass` (HomeAssistant) contains `asyncio.Future` objects that cannot be pickled.

The previous fix in commit `8485c91` used `get_context('fork')` to force the old method. This PR takes the proper approach: **remove `hass` and the `pygtfs.Schedule` object from `Process` args entirely**, so no pickling is required regardless of start method.

**Changes**:
- `get_gtfs()`: pass `joined_path` string instead of `hass`/`gtfs` to `Process`
- `extract_from_zip()`: reconstruct `pygtfs.Schedule` inside the child process from the path string; remove `hass` parameter
- `check_datasource_index()`: remove `hass` parameter; accept absolute `gtfs_dir`
- `check_extracting()`: remove `hass` parameter; callers now resolve `hass.config.path()` before calling
- All callers in `coordinator.py` and `config_flow.py` updated accordingly

### Issue 2 — SQLAlchemy connection/engine fd leak (2000+ open fds to TEC.sqlite)

**Root cause**: Two compounding leaks:

1. Each coordinator update cycle (every minute per sensor) called `get_gtfs()` → `pygtfs.Schedule(path)` → new SQLAlchemy engine + connection pool, but **never disposed the previous one**. Added `engine.dispose()` before reassigning `self._pygtfs` in both coordinator classes.

2. All `schedule.engine.connect().execute(...)` calls throughout `gtfs_helper.py` were made **without a context manager**, leaving connections open in the pool indefinitely. Wrapped all call sites with `with schedule.engine.connect() as conn:` and `.fetchall()` to eagerly consume results before the connection closes.

3. Added `schedule.engine.dispose()` at the end of `check_datasource_index` since it runs in a child process and should release resources before exit.

**Why visible now**: Python 3.14 changed GC timing (introduced incremental GC), so SQLAlchemy engine objects are no longer finalized as promptly as before, making the pre-existing leak accumulate rapidly.

### DIFF
--- a/custom_components/gtfs2/config_flow.py
+++ b/custom_components/gtfs2/config_flow.py
@@ -433,7 +433,7 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
         if self._pygtfs in ['no_data_file', 'no_zip_file', 'extracting'] :
             return self._pygtfs
         check_index = await self.hass.async_add_executor_job(
-                    check_datasource_index, self.hass, self._pygtfs, DEFAULT_PATH, data["file"]
+                    check_datasource_index, self._pygtfs, self.hass.config.path(DEFAULT_PATH), data["file"]
                 )            
         return None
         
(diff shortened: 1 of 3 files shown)


dafal:
Closing to test locally before merging.

PR #165, opened 2026-07-26
Derive departure dates from calendar_date for calendar_dates-only feeds

### LINKED ISSUE #164: [Bug]: tomorrow's trips filed under today for feeds with only calendar_dates.txt
### Own Analysis

```shell

For a start/end route on a feed that ships calendar_dates.txt without
calendar.txt, tomorrow's departures get listed under today.

get_next_departure() decides whether a row belongs to today or tomorrow by
comparing origin_depart_date, which is date(origin_stop_time.departure_time)
(lines 109 and 164). departure_time carries no date component, so SQLite
returns 1970-01-01 for every row, and 1970-01-02 for departures at or past
24:00:00. The comparison at lines 245-248 therefore can never distinguish two
service days. The code already knows the value is degenerate: line 367
compares it against the literal '1970-01-01'.

In the calendar_dates half of the UNION every row has exception_type = 1, so
row["today_cd"] == 1 is true for tomorrow's rows as well and they enter the
today branch. Because the query orders by calendar_date, tomorrow's rows then
overwrite today's at the same departure time.

What it does vs. expectation:
- on a normal day the two timetables share the same clock times, so the wrong
  day is invisible, but today's entries end up carrying tomorrow's trip_id and
  headsign
- on a day whose timetable differs from the next day's, the listed times are
  wrong
- on a day with no service at all, every listed departure is tomorrow's

Have you verified the source/provider is working correctly: yes. Metrolinx
correctly publishes zero trips for the affected dates. UP Express was fully
suspended on 2026-07-25 and 2026-07-26 for GO Expansion construction, and the
feed reflects that. The sensor still showed a full timetable for those days,
which is what led me here.
```

```shell

- re-downloaded 2026-07-26, feed_info feed_start_date 20260707, feed_end_date 20270103
- contains calendar_dates.txt and no calendar.txt, which is valid GTFS when
  calendar_dates covers all service dates
- service_id is the date itself, one service_id per date, all exception_type 1
- also reproduces with the GO Transit feed, same publisher, same structure:
  https://assets.metrolinx.com/raw/upload/Documents/Metrolinx/Open%20Data/GO-GTFS.zip

Note pygtfs synthesises 181 calendar rows from calendar_dates for this feed,
all with every weekday flag set to 0 and start_date = end_date = the single
date. So the calendar half of the UNION contributes no timetable entries here
and everything comes from the calendar_dates half.
```

```shell
direction: 0
include_tomorrow: true
offset: 5
```

```shell
n/a, this is a start/end route
```

```shell
gtfs2: 0.5.9
HA: HAOS 2026.7.2
```

### Relevant log output

```shell
My logger was at default: warning, so I have no debug output to paste. Happy to
capture it if useful. In the meantime here is the reproduction, run against the
real feed by replaying the calendar_dates half of the query with an injectable
"today" and then running the timetable loop from get_next_departure() over the
result.

simulated today   rows fetched   listed for today   actually tomorrow's
2026-07-26                  79                 74                    74   <- no service this day
2026-07-28                 158                 74                    74   <- normal day
2026-07-31                 154                 74                    70   <- Fri into Sat, timetables differ

On 2026-07-26 the feed has 0 Weston -> Union trips and 2026-07-27 has 79. The
sensor listed 05:39, 05:54, 06:06, 06:21 and so on, which are exactly
2026-07-27's departures.

Deriving the day from calendar_date, which those rows already carry, instead of
from origin_depart_date brings all three cases to 0 misattributed and leaves the
normal-day entry count unchanged at 153. PR to follow.

Possibly related but separate, so I have not bundled it: the date filters in
get_next_departure() use bare date('now'), which SQLite evaluates in UTC, while
the Python side computes now_date and the idx prefixes from local time via
dt_util.now(). West of UTC that shifts the queried window for part of each
evening. get_local_stops_next_departures() already binds date(:now_offset)
instead. I will open that separately.
```

ggiesen:
Fixes #164

Independent of #167. That one changes which days the query asks for, this one changes how the returned rows are placed on a calendar. They merge cleanly in either order.

### Problem

`get_next_departure()` decided which service day a row belonged to by comparing `origin_depart_date`, which is `date(origin_stop_time.departure_time)`. `departure_time` carries no date component, so pygtfs stores it on the epoch and that value is `1970-01-01` for every ordinary row. The comparison cannot tell two service days apart. The code already treated the value as degenerate: further down it was tested against the literal `'1970-01-01'`.

For feeds shipping `calendar_dates.txt` without `calendar.txt`, every row in the `calendar_dates` half of the UNION has `exception_type = 1`, so tomorrow's rows also satisfied the today test and were keyed with today's date. Ordering by `calendar_date` then let them overwrite today's rows at the same departure time.

### Fix

Those rows carry their service date in `calendar_date`, so use it.

GTFS times at or past `24:00:00` are normalised by pygtfs onto `1970-01-02`, so the distance from the epoch gives the number of days to add. That places a trip scheduled as `24:06` on the day *after* its service date rather than on the service date itself. The UP Express feed has 9079 stop_times at or past `24:00:00`, so this is routine rather than an edge case.

The same value then drives the state's own date, replacing this test:

```python
if _tomorrow and now_time > item['origin_depart_time']:
```

which reported today for anything still due later by the clock, even when the trip belonged to a different service date. That is why the state could disagree with `next_departures`.

Rows from the `calendar` half select `"" as calendar_date` and are untouched, so feeds with a real `calendar.txt` keep their existing behaviour.

### Verification

Against the Metrolinx UP Express feed, which has no `calendar.txt`. Every generated timetable key is checked against the departures actually present in the feed, with `24:00:00`+ rollover applied, and the selected next departure is compared to the true next one:

| scenario | before | after |
|---|---|---|
| Sunday, line suspended, 03:14 | 153 keys, **79 not in the feed**, next wrong by a day | 79 keys, 0 not in the feed, next correct |
| normal Tuesday, 03:14 | 153 keys, next correct | 158 keys, next correct |
| normal Tuesday, 23:50 | 153 keys, next correct | 158 keys, next correct |
| Friday into Saturday, 12:00 | 149 keys, next correct | 154 keys, next correct |

The extra keys on unaffected days are departures the old keying silently lost: after-midnight trips collided with same-clock-time trips on the adjacent day, so one overwrote the other.

### History of this PR

The first version was a two-line guard that only stopped tomorrow's rows entering the today bucket. Running it on my own instance showed that was not enough: it removed the bogus today-dated entries, but the after-midnight trips were still dated a day early, which made a trip actually running Tuesday 00:06 sort ahead of Monday 05:39 and become the reported next departure. The state also kept disagreeing with the attributes because of the separate clock-time test above. Both come from the same habit of inferring the date, so this replaces the guard with the date derivation rather than patching around it.

### DIFF
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -42,6 +42,36 @@ from .gtfs_rt_helper import get_rt_route_trip_statuses, get_gtfs_rt
 _LOGGER = logging.getLogger(__name__)
 
 
+def _service_departure_date(calendar_date, origin_depart_date):
+    """Return the date a departure actually falls on, as a YYYY-MM-DD string.
+
+    calendar_date is the service date taken from calendar_dates.txt, so it is
+    known rather than inferred. departure_time carries no date component, so
+    date() of it returns 1970-01-01, or 1970-01-02 for GTFS times at or past
+    24:00:00 which pygtfs normalises onto the following day. The distance from
+    the epoch is therefore the number of days to add to the service date, which
+    is what puts a trip scheduled as 24:06 on the day after its service date.
+
+    Returns None if either value cannot be parsed, so callers can fall back.
+    """
+    try:
+        service_date = datetime.datetime.strptime(str(calendar_date), "%Y-%m-%d")
+    except (TypeError, ValueError):
+        return None
+    try:
+        day_offset = (
+            datetime.datetime.strptime(str(origin_depart_date), "%Y-%m-%d")
+            - datetime.datetime(1970, 1, 1)
+        ).days
+    except (TypeError, ValueError):
+        day_offset = 0
+    if day_offset < 0:
+        day_offset = 0
+    return (service_date + datetime.timedelta(days=day_offset)).strftime(
+        dt_util.DATE_STR_FORMAT
+    )
+
+
 def get_next_departure(hass, _data):
     _LOGGER.debug("Get next departure with data: %s", _data)
     if check_extracting(hass, _data['gtfs_dir'],_data['file']):
@@ -229,6 +259,38 @@ def get_next_departure(hass, _data):
         
     for row_cursor in rows:
         row = row_cursor._asdict()
+        # Rows from the calendar_dates part of the query know their own service
+        # date, so place them directly instead of inferring the day from
+        # origin_depart_date. That value is date() of a time-only departure_time
+        # and is always the epoch, so it cannot tell one service day from another:
+        # without this, tomorrow's trips are filed under today, and because the
+        # query orders by calendar_date they overwrite today's at the same time.
+        departure_date = None
+        if row["calendar_date"]:
+            departure_date = _service_departure_date(
+                row["calendar_date"], row["origin_depart_date"]
+            )
+        if departure_date is not None:
+            is_today = departure_date == now_date_local_tz
+            # first and last keep the sentinels the day specific blocks below use:
+            # sensor.py drops the attribute when the value is None, so last stays
+            # None for anything that is not today, matching the tomorrow block.
+            extras = {
+                "day": "today" if is_today else "tomorrow",
+                "first": False,
+                "last": False if is_today else None,
+            }
+            idx = f"{departure_date} {row['origin_depart_time']}"
+            timetable[idx] = {**row, **extras}
+            if is_today:
+                if today_start is None:
+                    today_start = departure_date
+                    timetable[idx]["first"] = True
+                today_last = idx
+            elif tomorrow_start is None:
+                tomorrow_start = departure_date
+                timetable[idx]["first"] = True
+            continue
         if row["yesterday"] == 1 and yesterday_date >= row["start_date"]:
             extras = {"day": "yesterday", "first": None, "last": False}
             if yesterday_start is None:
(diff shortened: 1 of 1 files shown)


ggiesen:
Force-pushed a rewrite of this after dogfooding it on my own instance, so the diff is bigger than what you may have seen first.

The original was a two-line guard stopping tomorrow's rows from entering the today bucket. That removed the bogus today-dated departures, but two things were still wrong:

1. Trips scheduled at `24:00:00` or later were dated a day early. pygtfs normalises those onto `1970-01-02`, and the old keying ignored that, so a UP Express trip actually running Tue 00:06 sorted ahead of Mon 05:39 and became the reported next departure. That also means the previous keying was silently losing departures on ordinary days, where an after-midnight trip collided with a same-clock-time trip on the adjacent day: 153 keys before against 158 after, on a day where nothing else changed.
2. The state kept disagreeing with `next_departures`, because its date came from `if _tomorrow and now_time > item[origin_depart_time]`, which reports today for anything still due later by the clock even when the trip belongs to another service date.

Both are the same habit of inferring the date rather than using the one the feed supplies, so the guard is replaced with deriving the date from `calendar_date` plus the epoch offset. Feeds with a real `calendar.txt` are untouched, since those rows select `"" as calendar_date`.

Happy to split the state-formatting part into its own PR if you would rather keep them separate.

ggiesen:
Force-pushed a correction. I went back over this specifically looking for places where it disturbs behaviour it was not meant to touch, and found one I had got wrong.

`sensor.py` treats `None` as a sentinel meaning "this attribute does not apply here":

```python
if self._departure[ATTR_FIRST] is not None:
    self._attributes[ATTR_FIRST] = self._departure["first"]
elif ATTR_FIRST in self._attributes:
    del self._attributes[ATTR_FIRST]
```

The day specific blocks rely on that: the yesterday block sets `first: None`, the tomorrow block sets `last: None`, so those attributes are dropped rather than reported as false. My replacement set `first: False, last: False` for everything, so on a `calendar_dates` feed a tomorrow item started exposing `last: false` where the attribute had previously been absent. Anything templating on `is defined` would have changed behaviour.

Now mirrored: `last` stays `None` for anything that is not today.

While I was at it, the other things I checked and believe are unaffected:

- the `yesterday` block never applied to these rows anyway, since that half of the UNION selects `'0' AS yesterday`
- feeds with a real `calendar.txt` are untouched, because those rows select `"" as calendar_date` and the new path is guarded on that being non-empty
- both `first` and `last` keys are always present, so `sensor.py` cannot hit a `KeyError`

One thing I should be explicit about, because the PR title undersells it: this is **not** limited to feeds that lack `calendar.txt`. For a feed that has both, the `calendar_dates` half of the UNION still supplies the `exception_type = 1` added-service rows, and those now get dated from `calendar_date` too. I believe that is correct rather than merely different, since it is the same guess being replaced with the date the feed states, but it does mean hybrid feeds see a change and you may want to weigh that.

vingerha:
rejecting as not the right way, AI is not providing a generic solution and/or overly complex since it has no clue what the case is nor how this works in other gtfs areas

PR #167, opened 2026-07-26
Use the local date, not UTC, for the date filters in get_next_departure

### LINKED ISSUE #166: [Bug]: get_next_departure mixes UTC date('now') with local dates, shifting the day each evening west of UTC
### Own Analysis

```shell

get_next_departure() mixes two different notions of "today" in one query.

The SQL side uses bare date('now'), which SQLite always evaluates in UTC:

  line  90  AND (calendar_date_today.date = date('now'))
  line  98  AND (calendar_date_today.date = date('now') or
                 calendar_date_today.date = date('now','+1 day'))
  line  99  CASE WHEN date('now') < calendar_date_today.date THEN 1 else 0 END as tomorrow
  line 128  ... and date = date('now') and exception_type = 2 ...
  line 153  AND calendar.start_date <= date('now')
  line 154  AND calendar.end_date >= date('now')
  lines 185-186  date('now') AS start_date / AS end_date

The Python side derives everything from local time instead, via dt_util.now()
at lines 73-83, and those local values become the timetable keys:

  now_date, now_date_local_tz, tomorrow_date, tomorrow_date_local_tz

For any timezone behind UTC the two disagree for part of every evening. In
America/Toronto (UTC-4 in summer) from 20:00 local until midnight, date('now')
has already rolled over to tomorrow, so:

- the rows selected are tomorrow's and the day after's
- local today is never queried at all
- the rows that are selected get keyed with local today's date

What it does vs. expectation: during that window the departures shown for
"today" belong to a different service day. Expected is that the queried window
and the labels both refer to the same day, in the user's local timezone.

Have you verified the source/provider is working correctly: yes, this is purely
in the date arithmetic, independent of the feed contents.

Note there is already a precedent for the fix in this same file.
get_local_stops_next_departures() binds a Python-computed value instead of
using date('now'):

  line 901  AND (calendar_date_today.date = date(:now_offset))
  line 912  AND (calendar_date_today.date = date(:now_offset) or
                 calendar_date_today.date = date(:now_offset,'+1 day'))
  line 913  CASE WHEN date(:now_offset) < calendar_date_today.date ...
  lines 941-942  AND calendar.start_date <= date(:now_offset) ...

So aligning get_next_departure() with that pattern looks like the natural fix,
rather than reaching for date('now','localtime'), which would depend on the
container's TZ rather than on the HA configured timezone.

I have not sent a PR for this one because it also affects the calendar
start_date/end_date bounds, the exception_type = 2 subquery and the yesterday
handling, so I would rather hear which approach you prefer first.

Unrelated to the date arithmetic, but noticed while reading: const.py:291
defines CONF_SOURCE_TIMEZONE_CORRECTION = "source_timezone_correction" and
nothing references it anywhere else in the component. The option is stored in
config entries but has no effect. Happy to split that into its own issue if
you would prefer.
```

```shell

- re-downloaded 2026-07-26, feed_start_date 20260707, feed_end_date 20270103
- not feed specific though, this is in the date arithmetic and applies to any
  source when HA runs in a timezone behind UTC
```

```shell
direction: 0
include_tomorrow: true
offset: 5

HA timezone: America/Toronto (UTC-4 at time of writing)
```

```shell
n/a, this is a start/end route
```

```shell
gtfs2: 0.5.9
HA: HAOS 2026.7.2
```

### Relevant log output

```shell
My logger was at default: warning, so no debug output to paste. Happy to
capture it if useful.

Reproduction, modelling 21:00 EDT on Sunday 2026-07-26, when SQLite's
date('now') is already 2026-07-27 while the local date is still 2026-07-26:

  SQL date('now')      = 2026-07-27
  local today          = 2026-07-26
  service dates fetched by the query = {2026-07-27: 79, 2026-07-28: 79}
  was local today (2026-07-26) fetched at all?  no

Every resulting entry is then keyed with 2026-07-26 while holding rows for
2026-07-27 or 2026-07-28.

For clarity this is separate from #164 / #165, which is about how rows from the
calendar_dates half of the UNION are attributed to a day. That fix makes the
attribution self-consistent, but the whole window is still displaced during the
evening because the selection is UTC based and the labels are local. Both are
needed.
```

ggiesen:
Fixes #166

Independent of #165. That one changes how rows are attributed to a day in the timetable loop, this one changes which days the query asks for. They touch different lines and can be merged in either order.

### Problem

`get_next_departure()` mixes two notions of "today" in a single query.

The SQL side used bare `date('now')`, which SQLite always evaluates in UTC. The Python side derives `now_date`, `tomorrow_date` and the timetable keys from local time via `dt_util.now()` at lines 73-83. For any timezone behind UTC the two disagree for part of every evening: the queried window has already advanced a day while the labels have not, and the user's actual local day is never queried at all.

In `America/Toronto` at 21:00 local on Tue 2026-07-28, `date('now')` is `2026-07-29`.

### Fix

Bind the offset-adjusted local `now` and use `date(:now_offset)`, which is exactly what `get_local_stops_next_departures()` already does (`"now_offset": now` at line 976). Nine occurrences across eight lines, all inside `get_next_departure`.

`date('now','localtime')` was deliberately not used: it follows the container's `TZ` rather than the timezone configured in Home Assistant, so it would be right only by coincidence.

### Second change in the same lines

The `exception_type = 2` lookup for tomorrow interpolated `tomorrow_date` unquoted:

```python
... where service_id=trip.service_id and date = {tomorrow_date} and exception_type = 2
```

That renders as `date = 2026-07-29`, which SQLite parses as arithmetic and reduces to the integer `1990`, so the comparison can never match. A service removed for tomorrow via `exception_type = 2` was therefore still reported as running. The equivalent lookup for today, on the line below, was already quoted; this brings the two into line. Happy to split it into its own PR if you would rather keep them separate.

### Verification

Against the Metrolinx UP Express feed, which has service on both dates so a missing date can only be the filter's doing. `date('now')` cannot be simulated as it reads the real clock, so the current behaviour is modelled by pinning the literal date UTC would report at that moment.

Simulated local time 21:00 on 2026-07-28, expected window `2026-07-28` plus `2026-07-29`:

| | dates fetched | local today present | today flagged correctly |
|---|---|---|---|
| before | `2026-07-29`, `2026-07-30` | no | no, `2026-07-29` flagged as today |
| after | `2026-07-28`, `2026-07-29` | yes | yes |

During the day, when the local and UTC dates agree, `date(:now_offset)` and `date('now')` evaluate identically, so there is no change outside the offset window.

For the quoting fix, with an `exception_type = 2` row present for tomorrow: the old expression matched 0 rows, the new one matches 1.

The `datetime` bind was exercised through pygtfs's own SQLAlchemy engine rather than raw sqlite3, so it goes through the same path as the running integration.

### DIFF
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -87,16 +87,16 @@ def get_next_departure(hass, _data):
     # days.
     limit = 24 * 60 * 60 * 2
     tomorrow_select = tomorrow_select2 = tomorrow_where = tomorrow_order = ""
-    tomorrow_calendar_date_where = f"AND (calendar_date_today.date = date('now'))"
+    tomorrow_calendar_date_where = f"AND (calendar_date_today.date = date(:now_offset))"
     if include_tomorrow:
         _LOGGER.debug("Includes Tomorrow")
         limit = int(limit / 2 * 3)
         tomorrow_name = tomorrow.strftime("%A").lower()
-        tomorrow_select = f"( select calendar.{tomorrow_name} - ( select case when (select '1' from calendar_dates where service_id=trip.service_id and date = {tomorrow_date} and exception_type = 2 ) == '1' then '1' else '0' end) ) as tomorrow,"
+        tomorrow_select = f"( select calendar.{tomorrow_name} - ( select case when (select '1' from calendar_dates where service_id=trip.service_id and date = date(:now_offset,'+1 day') and exception_type = 2 ) == '1' then '1' else '0' end) ) as tomorrow,"
         tomorrow_where = f"OR calendar.{tomorrow_name} = 1"
         tomorrow_order = f"calendar.{tomorrow_name} DESC,"
-        tomorrow_calendar_date_where = f"AND (calendar_date_today.date = date('now') or calendar_date_today.date = date('now','+1 day') )"
-        tomorrow_select2 = f"CASE WHEN date('now') < calendar_date_today.date THEN 1 else 0 END as tomorrow,"
+        tomorrow_calendar_date_where = f"AND (calendar_date_today.date = date(:now_offset) or calendar_date_today.date = date(:now_offset,'+1 day') )"
+        tomorrow_select2 = f"CASE WHEN date(:now_offset) < calendar_date_today.date THEN 1 else 0 END as tomorrow,"
     sql_query = f"""
         SELECT trip.trip_id, trip.route_id,trip.trip_headsign, trip.direction_id,trip.trip_short_name,
                route.route_long_name,route.route_short_name,
@@ -125,7 +125,7 @@ def get_next_departure(hass, _data):
                destination_stop_time.stop_sequence AS dest_stop_sequence,
                destination_stop_time.timepoint AS dest_stop_timepoint,
                calendar.{yesterday.strftime("%A").lower()} AS yesterday,
-               ( select calendar.{now.strftime("%A").lower()} - (  select case when (select '1' from calendar_dates where service_id=trip.service_id and date = date('now') and exception_type = 2 ) == '1' then '1' else '0' end  ) ) as today,
+               ( select calendar.{now.strftime("%A").lower()} - (  select case when (select '1' from calendar_dates where service_id=trip.service_id and date = date(:now_offset) and exception_type = 2 ) == '1' then '1' else '0' end  ) ) as today,
                {tomorrow_select}
                calendar.start_date AS start_date,
                calendar.end_date AS end_date,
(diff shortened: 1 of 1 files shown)


ggiesen:
Force-pushed a small correction: the parameter is now bound as a date string rather than as a `datetime`.

I had bound `now` to match `get_local_stops_next_departures()`, but passing a `datetime` through `text()` to sqlite3 relies on its default adapter, which is deprecated as of Python 3.12. Confirmed against pygtfs's own engine:

```
datetime object  -> ('2026-07-28', '2026-07-29')   DeprecationWarning: the default datetime adapter is deprecated
date string      -> ('2026-07-28', '2026-07-29')   no warning
```

Same result, no warning, and one less thing to break when that adapter is eventually removed. Every use here is wrapped in `date()`, so a date string is all that is needed. Worth noting the existing call in `get_local_stops_next_departures()` does need a real `datetime`, because it also does `datetime(:now_offset, :timerange)` for the time range window, so that one cannot simply be changed the same way.

vingerha:
already resolved alternatively and in a way that I can recall to be in line with how I look at things, it may not be 100% ok but I am not acceptin gthis code

Issue #163, opened 2026-07-24
[Bug]: string "tomorrow" in the realtime-related fields causing ETA issue after midnight


Smurfjet:
### Own Analysis

- [x] I cannot find my issue in the documentation
- [x] My static data (zip) is up to date
- [x] The issue is not resolved with the latest (beta) release

### A clear and concise description of what the bug is.

```shell
Description:
When using Local Stop sensors, the attribute next_departures_lines frequently contains the string "tomorrow" in the realtime-related fields:

departure_realtime
departure_realtime_datetime
delay_realtime_derived
delay_realtime

This breaks any template, card, or sorting logic that expects a time, delay value, or null/"-".

To Reproduce:
Create a Local Stop sensor (based on a zone or person) with realtime enabled.
Observe the sensor attributes, especially around midnight or when trips from the next calendar day appear.
Look at entries in next_departures_lines that have "date" set to the next day.

Expected behavior:
Realtime fields should contain either:

A real departure time / delay, or
"-" / null (consistent with the “today” path when no realtime data is available)

They should never contain the literal string "tomorrow".

Actual behavior:
The integration hardcodes the string "tomorrow" into those four fields for next-day trips, so you see tomorrow for all the RT of busses scheduled after midnight.

I htink the issue is here in custom_components/gtfs2/gtfs_helper.py → function get_local_stops_next_departures:
Pythonif (row["tomorrow"] == '1' or row["tomorrow"] == 1) and \
   (datetime.datetime.strptime(now_time_hist_corrected,"%H:%M") > 
    datetime.datetime.strptime(row["departure_time"],"%H:%M:%S")):

    element = {
        ...
        "departure_realtime": "tomorrow",
        "departure_realtime_datetime": "tomorrow",
        "delay_realtime_derived": "tomorrow",
        "delay_realtime": "tomorrow",
        "date": tomorrow_date,
        ...
    }

I did not validate in detail, but the freindly local AI suggests:

if (row["tomorrow"] == '1' or row["tomorrow"] == 1) and \
   (datetime.datetime.strptime(now_time_hist_corrected, "%H:%M") > 
    datetime.datetime.strptime(row["departure_time"], "%H:%M:%S")):

    _LOGGER.debug("Tomorrow: adding row")
    element = {
        "departure": self._departure_time,
        "departure_datetime": self._departure_datetime_utc,
        "departure_realtime": "-",                      # ← changed
        "departure_realtime_datetime": "-",             # ← changed
[...]

PR #169, opened 2026-07-26
Stage datasource downloads and keep existing data when a refresh fails

### LINKED ISSUE #168: [Bug]: a failed datasource refresh deletes the existing data, and any response body is stored as the zip
### Own Analysis

```shell

get_gtfs() removes the existing zip and sqlite before fetching a replacement,
and then writes whatever comes back straight to the datasource without checking
the response:

  if update and data["extract_from"] == "url" and os.path.exists(...file...):
      remove_datasource(hass, path, filename, True)     # deletes zip AND sqlite
  ...
  if data["extract_from"] == "url":
      if not os.path.exists(os.path.join(gtfs_dir, file)):
          try:
              r = requests.get(url, headers=_headers, allow_redirects=True)
              open(os.path.join(gtfs_dir, file), "wb").write(r.content)
          except Exception as ex:
              _LOGGER.error("The given URL or GTFS data file/folder was not found")
              return "no_data_file"

Three things follow from that.

1. A failed refresh leaves nothing behind. Because remove_datasource() runs
   first, a network error means the datasource has neither its archive nor its
   database until the next successful run. Anyone calling update_gtfs from a
   weekly automation loses a week of departures to one transient failure.

2. Any response body becomes the archive. There is no raise_for_status(), so an
   error page is written as <file>.zip and the function returns normally, as
   though the refresh had worked. Extraction then fails on something that was
   never a zip. Note the caller cannot tell: for an HTTP error the return value
   is indistinguishable from success.

3. The request has no timeout, so a connection that opens and then stalls can
   block indefinitely.

What it does vs. expectation: a refresh that cannot complete should leave the
working datasource alone, and a non-2xx or non-zip response should be reported
rather than stored.

Have you verified the source/provider is working correctly: yes. Metrolinx
serves the feeds correctly. I hit this while setting up a scheduled refresh and
reasoning about what happens when it fails, then reproduced each mode directly.

Separate but adjacent, and I have not touched it: there is no conditional
request support. update_gtfs always re-downloads the full archive and always
re-extracts, with no ETag or Last-Modified handling, and no feed_version
comparison. For the GO Transit feed that is a 14.7MB download and a 227MB sqlite
rebuild taking roughly 12 minutes on a Home Assistant Yellow, repeated on every
call even when nothing has changed. Both Metrolinx endpoints do emit strong
ETags and Last-Modified and honour If-None-Match and If-Modified-Since with a
304, so skipping the work when the feed is unchanged is possible. Happy to open
that as a feature request separately if useful; the staging structure in the PR
makes it a natural follow-on.
```

```shell

Not feed specific, this is in the download path and applies to any url
datasource. The URLs used to reproduce the failure modes:

- HTTP 404 with an empty body:
  https://assets.metrolinx.com/raw/upload/Documents/Metrolinx/Open%20Data/GO-GTFS-Static.zip
- HTTP 200 with an HTML body:
  https://www.metrolinx.com/en/about-us/open-data
- unroutable host, to model a stalled connection:
  http://192.0.2.1/UP-GTFS.zip   (TEST-NET-1, RFC 5737)
```

```shell

Not config dependent. Triggered via:

  action: gtfs2.update_gtfs
  data:
    extract_from: url
    file: UP Express
    url: https://assets.metrolinx.com/raw/upload/Documents/Metrolinx/Open%20Data/UP-GTFS.zip
    api_key_name: api_key
    api_key_location: not_applicable
    clean_feed_info: false
    check_source_dates: false
```

```shell
n/a
```

```shell
gtfs2: 0.5.9
HA: HAOS 2026.7.2
```

### Relevant log output

```shell
My logger was at default: warning. Rather than logs, here is each failure mode
run directly against the live endpoints, starting from a working datasource
every time (a valid zip plus a sqlite) and reporting what survived:

  case                 result              datasource afterwards
  success, real feed   returned ok         valid zip, 901646 bytes, new data
  HTTP 404             returned ok    (!)  0 byte file saved as the zip, sqlite deleted
  HTTP 200 with HTML   returned ok    (!)  278445 bytes of HTML saved as the zip,
                                           sqlite deleted
  unroutable host      no_data_file        zip and sqlite both gone, nothing left

The two marked (!) are the ones I would highlight: the function reports success
while having destroyed the datasource, so a caller has no way to know the
refresh failed.

With the change in the PR, all three failure cases return no_data_file and leave
the existing zip and sqlite untouched.
```

ggiesen:
Fixes #168

Independent of #165 and #167. Those are both in `get_next_departure()`, this is in `get_gtfs()`. All three merge cleanly in any order.

### Problem

`get_gtfs()` removed the zip and the sqlite before fetching a replacement, then wrote whatever came back straight to the datasource without checking the response.

- **A failed refresh left nothing behind.** `remove_datasource()` ran first, so a network error meant the datasource had neither its archive nor its database until the next successful run.
- **Any response body became the archive.** No `raise_for_status()`, so an error page was saved as `<file>.zip` and the function returned normally, as though the refresh had worked. The caller could not distinguish an HTTP error from success.
- **No timeout**, so a connection that opened and then stalled could block indefinitely.

### Fix

Fetch to `<file>_download.zip`, require a 2xx, confirm the result really is a zip, and only then remove the old files and move the new one into place.

- the staging name avoids `<file>_temp.zip`, which `check_extracting()` uses as its marker, and is opened with `"wb"` so a leftover from an interrupted run is truncated rather than reused
- the timeout is a `(connect, read)` pair; with `stream=True` the read half applies between chunks, so a slow but progressing download is not cut off
- `zipfile.is_zipfile` comes from the vendored `zip_file` module already imported here

Behaviour is otherwise unchanged: a missing archive is still fetched, an existing one is still left alone unless a service call asks for a refresh, and a successful refresh still clears the sqlite so the feed is re-extracted.

### Verification

Each mode run against live endpoints, starting from a working datasource every time (a valid zip plus a sqlite):

| case | before | after |
|---|---|---|
| success, real feed | `ok`, new data | `ok`, new data |
| HTTP 404, empty body | **`ok`**, 0-byte file saved as the zip, sqlite deleted | `no_data_file`, zip and sqlite intact |
| HTTP 200 with HTML | **`ok`**, 278445 bytes of HTML saved as the zip, sqlite deleted | `no_data_file`, zip and sqlite intact |
| unroutable host | `no_data_file`, zip and sqlite both gone | `no_data_file`, zip and sqlite intact |

The two bold rows are the ones worth noting: the function reported success while having destroyed the datasource.

URLs used: `GO-GTFS-Static.zip` on the assets host for the 404, `metrolinx.com/en/about-us/open-data` for HTML with a 200, and `192.0.2.1` (TEST-NET-1) for the stall.

### Not included

No conditional request support. `update_gtfs` still re-downloads and re-extracts unconditionally, which for the GO Transit feed is a 14.7MB download and a 227MB sqlite rebuild of roughly 12 minutes on a Yellow, even when the feed has not changed. Both Metrolinx endpoints emit strong ETags and `Last-Modified` and honour `If-None-Match` and `If-Modified-Since` with a 304, so this is avoidable. The staging structure here makes it a natural follow-on, but it needs somewhere to persist the validator between runs, which felt like your call rather than something to slip into a bug fix. Happy to open it separately.

### DIFF
--- a/custom_components/gtfs2/const.py
+++ b/custom_components/gtfs2/const.py
@@ -13,6 +13,10 @@ DEFAULT_LOCAL_STOP_RADIUS = 200
 DEFAULT_MAX_LOCAL_STOPS = 15
 
 DEFAULT_NAME = "GTFS Sensor2"
+# (connect, read) timeout for fetching a static datasource. The read timeout
+# applies between chunks rather than to the whole transfer, so a slow but
+# progressing download is not cut off.
+DEFAULT_DOWNLOAD_TIMEOUT = (10, 60)
 DEFAULT_PATH = "gtfs2"
 DEFAULT_PATH_GEOJSON = "www/gtfs2"
 DEFAULT_PATH_RT = "www/gtfs2"
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -27,7 +27,8 @@ from .const import (
     CONF_API_KEY_LOCATION,
     CONF_API_KEY_NAME,
     CONF_ACCEPT_HEADER_PB,
-    DEFAULT_LOCAL_STOP_TIMERANGE, 
+    DEFAULT_DOWNLOAD_TIMEOUT,
+    DEFAULT_LOCAL_STOP_TIMERANGE,
     DEFAULT_LOCAL_STOP_TIMERANGE_HISTORY,
     DEFAULT_LOCAL_STOP_RADIUS,
     DEFAULT_PATH_RT,
@@ -473,23 +474,53 @@ def get_gtfs(hass, path, data, update=False):
     if check_extracting(hass, gtfs_dir,filename) and not update :
         _LOGGER.debug("Cannot use this datasource as still unpacking: %s", filename)
         return "extracting"
-    if update and data["extract_from"] == "url" and os.path.exists(os.path.join(gtfs_dir, file)):
-        remove_datasource(hass, path, filename, True)
     if update and data["extract_from"] == "zip" and os.path.exists(os.path.join(gtfs_dir, file)) and os.path.exists(os.path.join(gtfs_dir, sqlite)):
-        os.remove(os.path.join(gtfs_dir, sqlite))      
+        os.remove(os.path.join(gtfs_dir, sqlite))
     if data["extract_from"] == "zip":
         if not os.path.exists(os.path.join(gtfs_dir, file)):
             _LOGGER.error("The given GTFS zipfile was not found")
             return "no_zip_file"
     if data["extract_from"] == "url":
-        if not os.path.exists(os.path.join(gtfs_dir, file)):
+        target = os.path.join(gtfs_dir, file)
+        # Fetch when there is nothing yet, or when a service call asks for a
+        # refresh. Download to a staging file and only replace the datasource once
+        # the new archive has arrived complete and is really a zip. Removing the
+        # old zip and sqlite up front, as this used to, means a failed, truncated
+        # or non-zip response leaves the datasource with nothing at all until the
+        # next successful run.
+        if update or not os.path.exists(target):
+            staged = os.path.join(gtfs_dir, filename + "_download.zip")
             try:
-                r = requests.get(url,headers=_headers, allow_redirects=True)
-                open(os.path.join(gtfs_dir, file), "wb").write(r.content)
+                with requests.get(
+                    url,
+                    headers=_headers,
+                    allow_redirects=True,
+                    timeout=DEFAULT_DOWNLOAD_TIMEOUT,
+                    stream=True,
+                ) as r:
+                    r.raise_for_status()
+                    with open(staged, "wb") as staged_file:
+                        for chunk in r.iter_content(chunk_size=65536):
+                            staged_file.write(chunk)
             except Exception as ex:  # pylint: disable=broad-except
-                _LOGGER.error("The given URL or GTFS data file/folder was not found")
-                return "no_data_file"                
-    
+                _LOGGER.error(
+                    "Could not download GTFS data from: %s, keeping the current data, error: %s",
+                    url,
+                    ex,
+                )
+                if os.path.exists(staged):
+                    os.remove(staged)
+                return "no_data_file"
+            if not zipfile.is_zipfile(staged):
+                _LOGGER.error(
+                    "Data downloaded from: %s is not a zip archive, keeping the current data",
+                    url,
+                )
+                os.remove(staged)
+                return "no_data_file"
+            remove_datasource(hass, path, filename, True)
+            os.replace(staged, target)
+
     # if update (servicecall) then check if new file does not only have future dates
     if check_source_dates:
         if update and not check_calendar_dates_from_zip(gtfs_dir, file):


ggiesen:
One thing I should flag against this PR, since it is a behaviour change I did not think through when I opened it.

`get_gtfs()` returns strings on failure, and `DataUpdateCoordinator` assigns the return straight into `"schedule"` with no type check, while `get_next_departure()` does `schedule = _data["schedule"]` and later `schedule.engine.connect()`. So a string return reaches SQLAlchemy as a `str`.

That hazard predates this PR, but the old code happened to avoid it in the HTTP error case: it wrote the error body as the zip, forked an extract, and returned `"extracting"`, which the `check_extracting()` guard then caught. My version returns `"no_data_file"`, which that guard does not catch, so it can reach `get_next_departure()` and raise `AttributeError` instead of failing quietly.

Reachability is narrow, since it needs the zip to be missing as well as the download to fail:

- during setup, `config_flow` already checks for `no_data_file` / `no_zip_file` / `extracting`, so that path aborts cleanly
- on a normal refresh the archive is present, and this PR specifically stops it being deleted before the replacement arrives

So in practice it needs `/config/gtfs2` to have been cleared out from under an existing entry, for example a partial restore. Still, it trades a quiet failure for a traceback, which is a real regression in that corner.

Two ways to close it, and I would rather you pick than guess:

1. a two line type guard at the top of `get_next_departure()` (and the local stops equivalent) returning `{}` when `schedule` is not a `Schedule`, which also fixes the pre-existing case
2. keep the failure path returning `"extracting"`, which preserves current behaviour exactly but is a bit of a lie about what happened

Happy to add either here, or leave this PR to the download handling and open the guard separately.

Also for completeness, two smaller notes on this change:

- the old archive and the staged download coexist briefly, so peak disk use is roughly double the feed size for the duration. Irrelevant for the Metrolinx feeds, could matter for a very large one.
- nothing marks the datasource busy during the download, since `check_extracting()` looks for `_temp.zip` and the journal. Two concurrent refreshes would both write `_download.zip`. That race exists today as well, where they would both write `<file>.zip`, so this is not worse, and readers are better off because the existing data stays in place throughout.

vingerha:
I leave ONE comment here for all other issues/PR.. only when I get a grip on the other open issue (local stops / tomorrow / rt ) then I will commence on these....when time permits. At times I am not happy to never have thought about test scripts, it is a pain in the but to do manual stuff

ggiesen:
Would you like some help implementing pytest unit tests for the existing functionality?

vingerha:
If you have the knowledge and time then yes I would really appreciate tests but let's wait till after the PR... I need to come up with scenario that always bug me or in other words...complex scenario that reduce my motivation to fix :)

ggiesen:
No rush, that sequencing makes sense.

I built a suite in the meantime, so it is there whenever you want it rather than something you have to wait on:

https://github.com/ggiesen/gtfs2/tree/tests/pytest-suite

Deliberately not a PR. It is a branch on my fork to look at and argue with, and I would rather rework it to fit how you want to work than drop it on you.

141 tests: 77 pass, 20 fail on purpose, and 44 are opt-in against real agency feeds. Each deliberate failure is a reproduction with its issue number in the docstring, so they are runnable bug reports rather than claims. They are marked expected so CI stays green, and if one starts passing the run goes red to say the registry needs updating.

The three areas you mentioned being stuck on are where it is deepest, which was not planned. It is just where the reported bugs are.

Realtime is probably the one that saves you the most. The tests build GTFS-RT protobuf feeds directly and serve them over `file://` URLs, which `get_gtfs_feed_entities` already supports. A delayed bus, a cancelled trip, a trip crossing midnight are each a few lines to construct, with no agency API key and no waiting for the situation to happen. That is the thing #163 has you asking a user to paste debug output for.

One finding you may want regardless of the tests. Of six real feeds pulled (SNCF, LE MET', MBTA, Metro-North, GO Transit, UP Express), four define service with `calendar_dates.txt` only, SNCF included. Two ship both. None uses `calendar.txt` alone. So that branch is the common path rather than an edge case, which is worth knowing before triaging anything that lives in it.

On scenarios, yes please, that is the useful direction. If you list the ones that drain your motivation I would rather encode those than guess. The clock is controllable in the suite, including SQLite's `date('now')`, so DST boundaries, UTC rollover and end-of-validity dates are reachable on demand instead of a few times a year.

vingerha:
rejecting as fixed alternatively

PR #172, opened 2026-08-19
Fix realtime departures announced too early at terminus/layover stops

Pulpyyyy:
## Symptom

On a start/end sensor whose origin stop is a **terminus with a layover**, the realtime departures (`next_departures_realtime` / `next_departure_realtime`) are announced **too early** — 7 to 11 minutes on the feed below. The static schedule (`next_departures`) is correct, and the opposite direction of the same line is unaffected.

Observed on TAO Orléans (trip updates: `https://ara-api.enroute.mobi/tao/gtfs/trip-updates`), line 40, direction Gare d'Orléans → Gaston Galloux, origin stop `ORLEANS:StopArea:01001712` (Gare d'Orléans - Quai E):

| shown by gtfs2 | actual departure |
|---|---|
| 14:32 | 14:43 |
| 15:03 | 15:13 |
| 15:35 | 15:42 |

## Cause

`get_rt_route_trip_statuses` keeps the **arrival** time of the `stop_time_update`, with the departure time only as fallback:

```python
# Use stop arrival time;
# fall back on departure time if not available
if stop["arrival"]["time"] == 0:
    stop_time = stop["departure"]["time"]
else:
    stop_time = stop["arrival"]["time"]
```

Mid-route, `arrival == departure`, so the choice makes no visible difference and the bug stays invisible. At a terminus or a timing/regulation point the vehicle stands still for a while and the two values diverge. Raw trace of the feed:

```
Line 40, dir=1, Quai E (01001712) — terminus of the return leg
  seq=1  arrival=14:32:41  departure=14:43:00   layover=10 min
  seq=1  arrival=15:03:13  departure=15:13:00   layover=9 min
  seq=1  arrival=15:35:00  departure=15:42:00   layover=7 min

Line 40, dir=0, Gaston Galloux (00026500) — mid-route stop
  seq=19 arrival=14:24:11  departure=14:24:11   identical
```

Both callers of this function consume the result as **departure** times:

- `gtfs_rt_helper.get_next_services` → key `"departures"`, exposed as `next_departures_realtime` / `next_departure_realtime`;
- the local-stops path in `gtfs_helper.get_local_stops_next_departures` → `departures` / `departure_rt`, compared against the planned **departure** time.

So the departure time is the correct value in both cases: the arrival time announces a bus that is standing at the platform but has not left yet.

(For context: line 40 is a loop, so the same stop_id appears in both directions at this stop. That was the first suspect, but replaying the entity-matching condition with the sensor's real values showed all retained entities were the correct `dir=1, seq=1` — the direction filter works. The gap came solely from the arrival/departure choice inside the right `stop_time_update`.)

## Fix

Invert the preference — departure first, arrival as fallback:

```python
# Use stop departure time;
# fall back on arrival time if not available
if stop["departure"]["time"] == 0:
    stop_time = stop["arrival"]["time"]
else:
    stop_time = stop["departure"]["time"]
```

The fallback remains necessary: some feeds only publish one of the two times.

## Verification

Old and new logic simulated on the live TAO feed, compared to the static schedule:

| sensor | before (arrival) | after (departure) | schedule |
|---|---|---|---|
| 40 Galloux→Gare | 14:24, 14:54, 15:24 | 14:24, 14:54, 15:24 | 14:24, 14:54, 15:25 |
| **40 Gare→Galloux** | 14:32, 15:03, 15:35 | **14:43, 15:13, 15:42** ✅ | 14:43, 15:13, 15:42 |
| 41 Galloux→Gare | 16:22, 16:43, 17:00 | 16:22, 16:43, 17:00 | 16:21, 16:43, 16:59 |
| **41 Gare→Galloux** | 16:15, 16:34, 16:56 | **16:15, 16:41, 17:06** ✅ | 16:15, 16:41, 17:06 |

The unaffected directions are byte-identical before/after (arrival == departure mid-route); their one-minute deviations from the schedule are legitimate realtime, not a regression.

## Deliberately not touched

The delay selection just below the changed block keeps `max(departure_delay, arrival_delay)`. For consistency it could prefer the departure delay, but the `max()` is a defensible fallback for feeds that only fill the arrival delay, and the observed feed publishes no `delay` field at all. Can be revisited separately if a feed with asymmetric delays shows up.

### DIFF
--- a/custom_components/gtfs2/gtfs_rt_helper.py
+++ b/custom_components/gtfs2/gtfs_rt_helper.py
@@ -281,12 +281,15 @@ def get_rt_route_trip_statuses(self):
                             departure_times[self._route_id][direction_id][stop_id]["departures"] = []
                             departure_times[self._route_id][direction_id][stop_id]["delays"] = []
                         
-                        # Use stop arrival time;
-                        # fall back on departure time if not available                            
-                        if stop["arrival"]["time"] == 0:
-                            stop_time = stop["departure"]["time"]
-                        else:
+                        # Use stop departure time;
+                        # fall back on arrival time if not available
+                        # both callers report departures from this stop, so the departure time is
+                        # the correct one: at a terminus or layover the vehicle can stand several
+                        # minutes, and the arrival time would announce a bus that has not left yet
+                        if stop["departure"]["time"] == 0:
                             stop_time = stop["arrival"]["time"]
+                        else:
+                            stop_time = stop["departure"]["time"]
                             
                         if stop["departure"].get("delay",0) >= stop["arrival"].get("delay",0):
                             delay = stop["departure"].get("delay",0)
--- a/custom_components/gtfs2/gtfs_rt_helper.py
+++ b/custom_components/gtfs2/gtfs_rt_helper.py
@@ -281,15 +281,15 @@ def get_rt_route_trip_statuses(self):
                             departure_times[self._route_id][direction_id][stop_id]["departures"] = []
                             departure_times[self._route_id][direction_id][stop_id]["delays"] = []
                         
-                        # Use stop departure time;
-                        # fall back on arrival time if not available
-                        # both callers report departures from this stop, so the departure time is
-                        # the correct one: at a terminus or layover the vehicle can stand several
-                        # minutes, and the arrival time would announce a bus that has not left yet
-                        if stop["departure"]["time"] == 0:
-                            stop_time = stop["arrival"]["time"]
-                        else:
-                            stop_time = stop["departure"]["time"]
+                        # Both callers report departures from this stop, so the
+                        # later of the two times is the one to announce: at a
+                        # terminus or a layover the vehicle stands several
+                        # minutes at its bay, and the arrival would announce a
+                        # bus that has not left yet. A time the feed does not
+                        # publish arrives as 0, so it can never be the later
+                        # one, which is the fallback for the poorer feeds.
+                        stop_time = max(stop["arrival"]["time"],
+                                        stop["departure"]["time"])
                             
                         if stop["departure"].get("delay",0) >= stop["arrival"].get("delay",0):
                             delay = stop["departure"].get("delay",0)
--- a/custom_components/gtfs2/gtfs_rt_helper.py
+++ b/custom_components/gtfs2/gtfs_rt_helper.py
@@ -281,13 +281,9 @@ def get_rt_route_trip_statuses(self):
                             departure_times[self._route_id][direction_id][stop_id]["departures"] = []
                             departure_times[self._route_id][direction_id][stop_id]["delays"] = []
                         
-                        # Both callers report departures from this stop, so the
-                        # later of the two times is the one to announce: at a
-                        # terminus or a layover the vehicle stands several
-                        # minutes at its bay, and the arrival would announce a
-                        # bus that has not left yet. A time the feed does not
-                        # publish arrives as 0, so it can never be the later
-                        # one, which is the fallback for the poorer feeds.
+                        # the later of the two 'time' attributes is the one to announce
+                        # e.g. at a terminus/layover where the vehicle stands several
+                        # minutes at its bay
                         stop_time = max(stop["arrival"]["time"],
                                         stop["departure"]["time"])
                             


vingerha:
Funny, it took a while for someone to come with this, already 3y ago I had questions about the use of arrival/departure time which I mostly copied from gtfs setup, assuming this to be some sort of standard.
Just for my understanding, why would a vehicle wait (roughly) 10 minutes before continuing its trip?

Pulpyyyy:
Layover at the terminus. Line 40 is a loop: Gare d'Orleans Quai E is both the last stop of direction 0 (seq 25) and the first stop of direction 1 (seq 1). Quai E is a bay of the Gare d'Orleans bus station: space is tight and maneuvers are restricted there, so the bus does not move off to a depot or a turnaround loop, it simply stays at its bay. It arrives, stands 7 to 11 minutes (driver break plus schedule regulation), then starts the next run from the same spot. The raw feed shows it clearly, in the same stop_time_update: arrival 14:32:41, departure 14:43:00.

Mid-route arrival equals departure, so the choice between the two is invisible everywhere except at a terminus or a timing point. That is probably why nobody hit it in 3 years.

And since both callers of `get_rt_route_trip_statuses` expose the result as departures (`next_departures_realtime`), the departure time is the semantically correct one. The fallback on arrival stays: some feeds publish only one of the two times.

For context: these four PRs come from the same work. I am building a custom Lovelace card on top of gtfs2, a departures panel plus an interactive line map that reads the GeoJSON files directly. That is where the fixes surfaced. Current state below.
image
image
Available also on mobile device (with many vehicules right now)
Screenshot_2026-08-20-10-56-48-076_io homeassistant companion android

vingerha:
in principle, the rt data is based on the trip_id, so if the bus/vehicle changes direction, it should (!) change trip_id and thus also rt data. I recently changed the code as there were irregularities in them but I am not sure if that would cover this one right now.
Since I see that this is a start/stop setup, can you verify the rt data for both directions, i.e. you should have setup two entities, and let me know if the trip_id is different? In quite a few cases in France (also with me in the south, sometimes there is no trip_id at all.

EDIT: this info may not help much, just thought that if the post-layover trip 'arrives' at the same stop then it is still not correct, but it would be nice to know. Additional question, does the rt data show both arrival and departure for the realtime values?

EDIT2: please reminisce with me... as mentioned I sort-of copied the current selection process from old gtfs. I am now thinking why it should not use the relatime departure_time for all cases, would you know of a case where this is not correct?

Pulpyyyy:
I pulled your questions against the live TAO feed (trip-updates, 17:20 today), line 40:

1. **Different trip_id per direction: yes.** At the Quai E stop right now: 9 trip_ids for dir 0, 8 for dir 1, none shared. The three buses in rotation (4001, 4002, 4003) all show the same pattern: the outbound trip (`40_A`) ends at seq 25, a new inbound trip_id (`40_R`) starts at seq 1, with a 7 to 10 min layover in between.

```
vehicle 4003:  dir=0 seq=25  arrival=17:37:31  departure=17:37:31  trip_id=ORLEANS:VehicleJourney:40_A_86_8_4003_4_165700
               dir=1 seq=1   arrival=17:37:31  departure=17:46:00  trip_id=ORLEANS:VehicleJourney:40_R_85_8_4003_4_174600
vehicle 4001:  dir=0 seq=25  arrival=18:08:05  departure=18:08:05  trip_id=ORLEANS:VehicleJourney:40_A_86_8_4001_4_172900
               dir=1 seq=1   arrival=18:08:00  departure=18:18:00  trip_id=ORLEANS:VehicleJourney:40_R_85_8_4001_4_181800
vehicle 4002:  dir=0 seq=25  arrival=18:41:01  departure=18:41:01  trip_id=ORLEANS:VehicleJourney:40_A_86_8_4002_4_180100
               dir=1 seq=1   arrival=18:41:01  departure=18:48:00  trip_id=ORLEANS:VehicleJourney:40_R_85_8_4002_4_184800
```

Note that the inbound trip's arrival is identical to the outbound trip's final arrival (17:37:31 twice for bus 4003): both records describe the same physical moment, the bus reaching its bay. The end of one trip and the start of the next are only told apart by the trip_id, and the actual restart time exists in one place only, the inbound departure field. That duplicated arrival is exactly what the old code was announcing as a departure, 9 minutes early. So the direction matching in gtfs2 is fine, I verified it separately: the wrong time never came from another trip, it came from inside the correct stop_time_update.

2. **Both times in the realtime data: yes**, all 26 updates in this snapshot carry arrival and departure. Following bus 4003 along its run:

```
mid-route (Gaston Galloux, seq=19):   arrival=17:27:54  departure=17:27:54   equal
end of trip (Quai E, seq=25):         arrival=17:37:31  departure=17:37:31   equal
next trip after layover (seq=1):      arrival=17:37:31  departure=17:46:00   9 min apart
```

The two times are equal everywhere except at the start of a trip after a layover, which is exactly where the old code picked the wrong one (the TAO trip_id even ends with the scheduled departure, `_174600`).

3. **Using the realtime departure everywhere:** correct whenever a departure is present, and I know of no case where preferring it is wrong. The arrival fallback is kept as a non-regression guarantee: feeds that publish only one of the two times, and the last stop of a trip where a departure makes no sense, keep behaving exactly as today. That is what the PR does: departure first, arrival as fallback when it is missing. So the simple rule you are thinking of is this patch, with a safety net for the poorer feeds.

vingerha:
Whilst on the train I started to remember a few rare csaes why 'arrival' is important too. In CH the vehicles will just pass-by the stop if there is no one waiting, even when being 1-2 mins early, and on certain stops they wait to recover/delay when being too early. I then thought to provide both arrival and departure as separate attributes so people can choose themselves what to use...will give this a further thought as I then should also apply this to the stop-based setup. Let me know of yor ideas or other thoughts :)

vingerha:
Re on above, as this is only realtime which I did not really notice with jet-lag before :), I withdraw the above idea. My current thought is now to use the time that is 'latest', so if departure time > arrival time then take that....this would also resolve any mid-route wait-time, e.g. at a bus terminal.

Pulpyyyy:
Fine for me using the max date

vingerha:
thx

Issue #170, opened 2026-08-06
[Bug]: Vehicle position found by trip_id but not added to GeoJSON when static GTFS has no direction_id (Renfe Cercanías)


masterfish1:
### Own Analysis

- [x] I cannot find my issue in the documentation
- [x] My static data (zip) is up to date
- [x] The issue is not resolved with the latest (beta) release

### A clear and concise description of what the bug is.

```shell
I'm using gtfs2 v0.5.9 with the official Renfe Cercanías GTFS and GTFS-RT feeds.

Realtime departures work correctly, but vehicle positions are never added to the generated GeoJSON. The GeoJSON file is created successfully but always contains:

{"features":[],"type":"FeatureCollection"}

As a result, no vehicles are displayed on the map.

I have verified that the Renfe data source is working correctly:

The official GTFS static feed contains the configured trip_id (4016J24070C2).
The official vehicle_positions.pb feed also contains the same trip_id.
The equivalent vehicle_positions.json feed contains the same tripId together with valid latitude, longitude and vehicle ID.

Therefore the realtime vehicle exists in the provider's feed, but gtfs2 does not add it to the GeoJSON.

This is my first time setting it up, so can't confirm whether it worked in previous versions.
```

### Provide details on the data source(s)

```shell
Static GTFS:
https://data.renfe.com/dataset/horarios-cercanias

Realtime feeds:
Trip Updates: https://gtfsrt.renfe.com/trip_updates.pb
Vehicle Positions: https://gtfsrt.renfe.com/vehicle_positions.pb
Alerts: https://gtfsrt.renfe.com/alerts.pb
```

### Config in case of setup using start / stop station

```shell
Origin station: Alzira
Destination station: València Nord
Route ID: 40T0006C2
Trip ID: 4016J24070C2
Origin stop ID: 64104
Direction: None (the Renfe static GTFS does not contain a direction_id column)
```

### Config in case of setup using location based departures

```shell

```

### Info on your installed versions in X.Y.Z format

gtfs2 version to be found in HACS top left alike 0.5.9
HA type (HAOS/Container) and release alike 2026.8.1

### Relevant log output

```shell
[36m2026-08-06 18:37:50.792 DEBUG (SyncWorker_27) [custom_components.gtfs2.gtfs_rt_helper] Configuration for RT route: 40T0006C2, RT trip: 4016J24070C2, RT stop: 64104, RT direction: None, trip short name: None[0m
[...]

Issue #148, opened 2026-01-09
[Bug]: Realtime feed for Pittsburgh Regional Transit not populating, all schedule_relationship fields are SCHEDULED


leinaht:
### Own Analysis

- [x] I cannot find my issue in the documentation
- [x] My static data (zip) is up to date
- [x] The issue is not resolved with the latest (beta) release

### A clear and concise description of what the bug is.

```shell
I think this is actually an issue with my data provider but wanted input before reaching out. Static data works, I can add the realtime feeds and see that the integration is calling the realtime feed but those attributes remain blank. Looking at the realtime trip updates feed, each schedule_relationship field shows as SCHEDULED - is that causing the integration to ignore them, and should there be a different status for realtime trip updates?
```

### Provide details on the data source(s)

```shell
static: https://www.rideprt.org/developerresources/GTFS.zip

realtime: https://truetime.portauthority.org/gtfsrt-bus/trips
```

### Config in case of setup using start / stop station

```shell
route ID: 58
stop ID: 18955
direction: inbound
```

### Config in case of setup using location based departures

```shell

```

### Info on your installed versions in X.Y.Z format

0.5.6.6
HAOS 2026.1.0 (VM QEMU)

### Relevant log output

```shell

```

vingerha:
I will have a look. It usually is the data provider but...I am not perfect too :)

The easiest way to test is to setup the RT extraction using the service/action and tick the debug switch, that will output the data in readable format and allows you to see if the trip or route of our static data (HA entity) exists in the RT file and/or has the correct format. Where I live, one (1) provider switched to changing data almost daily, meaning that I needed to update the static data similarly fast...they never properly explained why this is required/chosen :(

Issue #146, opened 2025-11-29
[Bug]: stale geojson entities


bojankoma:
### Own Analysis

- [x] I cannot find my issue in the documentation
- [x] My static data (zip) is up to date
- [x] The issue is not resolved with the latest (beta) release

### A clear and concise description of what the bug is.

```shell
I've been using this integration for weeks now and my only complaint is that it creates stale GeoJSON entities when used with Home Assistant GeoJSON integration. 

It's an easy fox that I've tested manually. For each property object in GeoJSON dict that you add 'id', you must add 'external_id' too. This property is exclusively used by GeoJSON integration as a unique ID. 
Tested this manually and locally and it works, but I'd like the change to persist over updates. 

Wanted to submit a PR fixing this but I have no permissions to do so, hence opening an  issue.
```

### Provide details on the data source(s)

```shell
GTFS Realtime Data
https://www.zet.hr/gtfs-rt-protobuf

GTFS Static Data
https://www.zet.hr/gtfs-scheduled/latest
```

### Config in case of setup using start / stop station

```shell
Any combination will suffice.
```

### Config in case of setup using location based departures

```shell

```

### Info on your installed versions in X.Y.Z format

gtfs2 version 0.5

HAOS
Core 2025.11.3
Supervisor 2025.11.5
Operating System 16.3
Frontend 20251105.1

### Relevant log output

```shell

```

vingerha:
Hi, thanks for this and tbh it has been quite a long time ago that I looked at it. I would a appreciate a PR so I can focus on the key thing that you discovered. Normally...you have to apply it on your own fork and then raise it to this repo...normally. Does that not work?

EDIT: alternatively add the code to this issue as a file (with a short description if multiple files/locations)

Issue #126, opened 2025-02-06
[Bug]: Random Errore on RealTime Data


andrea-pustina:
### Own Analysis

- [x] I cannot find my issue in the documentation

### Own Analysis

- [x] My static data (zip) is up to date

### A clear and concise description of what the bug is.

```shell
Error getting gtfs realtime data


I think that occurs when RealTime Data is empty.

When there is this error all sensor become unavailable. Maybe last update values should be retained
```

### Provide details on the data source(s)

```shell
static: rome_static_gtfs.zip
trip updates: https://romamobilita.it/sites/default/files/rome_rtgtfs_trip_updates_feed.pb
vehicles: https://romamobilita.it/sites/default/files/rome_rtgtfs_vehicle_positions_feed.pb
service alerts: https://romamobilita.it/sites/default/files/rome_rtgtfs_service_alerts_feed.pb
```

### Config in case of setup using start / stop station

```shell
route_id: 714
start_stop: 70243
end_stop: 70698
```

### Config in case of setup using location based departures

```shell

```

### Info on your installed version

Latest/haos

### Relevant log output

```shell
This error originated from a custom integration.

Logger: custom_components.gtfs2.coordinator
Source: custom_components/gtfs2/coordinator.py:157
integration: GTFS 2 (General Transit Feed Specification) (documentation, issues)
First occurred: 12:12:27 (36 occurrences)
Last logged: 17:09:16

Error getting gtfs realtime data, for origin: 70667: NAVIGATORI (14) with error: Error parsing message with type 'transit_realtime.FeedMessage'
Error getting gtfs realtime data, for origin: 70243: S. MARIA MAGGIORE (2) with error: Error parsing message with type 'transit_realtime.FeedMessage'
```

vingerha:
Which version of gtfs2,..... that was asked in the questions, latest is NOT an answer

vingerha:
![Image](https://github.com/user-attachments/assets/89b30d51-b546-439a-987e-d4a9987f9b73)

andrea-pustina:
0.5.5

vingerha:
Then please install the latest beta7, I applied some fixes the last week

vingerha:
BTW, those fixes will not fi issues with the realtime source, they are there to prevent that the compete entoty fails to be created on failed/missing/incorrect realtime data

vingerha:
And I just checked ... there are none of these stops on route 714

vingerha:
No feedback, closing

PR #173, opened 2026-08-19
Stable identity and readable labels for realtime vehicle position markers

Pulpyyyy:
## Two problems with the realtime vehicle position markers

`get_rt_vehicle_positions` builds `properties.id` and `properties.title` of the GeoJSON features from the raw `route_id`, the `direction_id` and the last 3 digits of the crc32 of the `trip_id`.

### 1. Unstable identity → unbounded entity-registry growth

The `geo_json_events` integration registers each marker with `unique_id = <config_entry_id>_<properties.id>`:

```
geo_location.orleans_line_40_0_156_bus
  unique_id = 01M0D312WVNW3VYW8CKQDFE0YE_ORLEANS:Line:40(0)156
```

The crc32 of the `trip_id` changes **on every trip**, so every vehicle comes back as a brand-new entity on each run, and the registry grows without bound. Observed on a live install (TAO Orléans feed), registry vs live states after only ~30 minutes:

```
in registry: 5        alive: 4
  ALIVE  geo_location.orleans_line_40_0_156_bus
  stale  geo_location.orleans_line_40_0_547_bus   <- bus left the feed
  ALIVE  geo_location.orleans_line_40_1_189_bus
  ALIVE  geo_location.orleans_line_40_1_720_bus
  ALIVE  geo_location.orleans_line_40_1_934_bus
```

### 2. Unreadable labels on the map

`geo_json_events` uses `properties.title` as the entity name, and the map card displays it as the marker label: `ORLEANS:Line:40(0)156_bus`. Worse, the marker badge only shows the first letter of each word (3 max, see `ha-map.ts`), and this title contains no space — so the badge shows a single letter.

## Fix

Build the identity from what is actually stable, and the label from what a human reads:

- `properties.id` = `<route_id>_<direction>_<vehicle id>` (e.g. `ORLEANS:Line:40_1_7015`). The feed's `vehicle.id` is stable across trips and days, so a vehicle keeps the same `unique_id`, hence the same entity and name. The registry stabilizes at the size of the fleet serving the tracked line.
- `properties.title` = `<line short name> → <destination> <vehicle id>` (e.g. `40 → Gaston Galloux 7015`). The short name comes from the schedule (`get_next_departure` now returns it next to `route_id`), the destination from the config entry, cleaned of the stop-id prefix, platform suffix and counter.

Fallbacks keep every previous behavior reachable:

- feed publishes no `vehicle.id` → fall back on the crc32 digits (original behavior for that vehicle);
- config entry has no usable destination (e.g. local-stop sensors) → fall back on the original full label;
- the destination parsing is wrapped in `try` and the vehicle descriptor is read with `.get()` — a label is cosmetic and must never break the update.

Before/after on the live feed:

```
before:  geo_location.orleans_line_40_1_934_bus   (new entity every trip)
after:   geo_location.40_gaston_galloux_7015      (stable across trips/days)
         friendly name: 40 → Gaston Galloux 7015
```

The registry entries left by the old ids are cleaned up in a separate PR (#179).

## Known limitation

If a feed stops publishing `vehicle.id`, the crc fallback brings back the original per-trip churn for that vehicle. On the observed feed all 94 vehicles carry one.

Verified live on TAO Orléans (`https://ara-api.enroute.mobi/tao/gtfs/vehicle-positions`): stable entity set (~a dozen for two lines), readable names on the map card, local-stop sensors untouched (fallback label).

### DIFF
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -505,6 +505,7 @@ def get_next_departure(hass, _data):
     data_returned = {
         "trip_id": item["trip_id"],
         "route_id": item["route_id"],
+        "route_short_name": item["route_short_name"],
         "trip_direction_id": item["direction_id"],
         "trip_short_name": item["trip_short_name"],
         "day": item["day"],
--- a/custom_components/gtfs2/gtfs_rt_helper.py
+++ b/custom_components/gtfs2/gtfs_rt_helper.py
@@ -344,8 +344,32 @@ def get_rt_vehicle_positions(self):
             geojson_element["geometry"]["coordinates"] = []
             geojson_element["geometry"]["coordinates"].append(vehicle["position"]["longitude"])
             geojson_element["geometry"]["coordinates"].append(vehicle["position"]["latitude"])
-            geojson_element["properties"]["id"] = str(self._route_id) + "(" + str(vehicle["trip"]["direction_id"]) + ")" + str(binascii.crc32((vehicle["trip"]["trip_id"]).encode('utf8')))[-3:]
-            geojson_element["properties"]["title"] = str(self._route_id) + "(" + str(vehicle["trip"]["direction_id"]) + ")" + str(binascii.crc32((vehicle["trip"]["trip_id"]).encode('utf8')))[-3:] + "_" + self._icon.split(':')[1]
+            # Stable, human-readable marker identity.
+            # geo_json_events registers each marker with unique_id
+            # <config_entry_id>_<properties.id>; the crc32 of the trip_id changes
+            # every trip, so each vehicle came back as a brand-new entity on every
+            # run and the entity registry grew without bound. The vehicle id is
+            # stable across trips and days, so the entity (and its name) is reused.
+            # The id joins the raw route_id, the direction and the vehicle id
+            # with underscores, so it cannot be mistaken for the old
+            # route_id(direction)crc form. The map card uses the title as
+            # entity name: line short name, destination, vehicle. Fall back on
+            # the original crc-based label when the feed publishes no vehicle
+            # id or the sensor carries no usable line name or destination.
+            _crc = str(binascii.crc32((vehicle["trip"]["trip_id"]).encode('utf8')))[-3:]
+            _veh = str(vehicle.get("vehicle", {}).get("id", "") or vehicle.get("vehicle", {}).get("label", "")).strip()
+            _dir = str(vehicle["trip"]["direction_id"])
+            try:
+                _line = str(self._data.get("next_departure", {}).get("route_short_name") or "").strip()
+                _dest = self.config_entry.data.get("destination", "").split(": ")[-1].split(" (")[0].split(" - ")[0].strip()
+            except Exception:  # the label is cosmetic, it must never break the update
+                _line, _dest = "", ""
+            if _line and _dest:
+                _label = _line + " → " + _dest + " " + (_veh or _crc)
+            else:
+                _label = str(self._route_id) + "(" + _dir + ")" + _crc + "_" + self._icon.split(':')[1]
+            geojson_element["properties"]["id"] = str(self._route_id) + "_" + _dir + "_" + (_veh or _crc)
+            geojson_element["properties"]["title"] = _label
             geojson_element["properties"]["trip_id"] = vehicle["trip"]["trip_id"]
             geojson_element["properties"]["route_id"] = str(self._route_id)
             geojson_element["properties"]["direction_id"] = vehicle["trip"]["direction_id"]
--- a/custom_components/gtfs2/gtfs_rt_helper.py
+++ b/custom_components/gtfs2/gtfs_rt_helper.py
@@ -344,25 +344,15 @@ def get_rt_vehicle_positions(self):
             geojson_element["geometry"]["coordinates"] = []
             geojson_element["geometry"]["coordinates"].append(vehicle["position"]["longitude"])
             geojson_element["geometry"]["coordinates"].append(vehicle["position"]["latitude"])
-            # Stable, human-readable marker identity.
-            # geo_json_events registers each marker with unique_id
-            # <config_entry_id>_<properties.id>; the crc32 of the trip_id changes
-            # every trip, so each vehicle came back as a brand-new entity on every
-            # run and the entity registry grew without bound. The vehicle id is
-            # stable across trips and days, so the entity (and its name) is reused.
-            # The id joins the raw route_id, the direction and the vehicle id
-            # with underscores, so it cannot be mistaken for the old
-            # route_id(direction)crc form. The map card uses the title as
-            # entity name: line short name, destination, vehicle. Fall back on
-            # the original crc-based label when the feed publishes no vehicle
-            # id or the sensor carries no usable line name or destination.
+            # Altered to use vehicle_id (if existing) to create the unique indicator instead of trip_id
+            # to reduce number of entities created by geojson. 
             _crc = str(binascii.crc32((vehicle["trip"]["trip_id"]).encode('utf8')))[-3:]
             _veh = str(vehicle.get("vehicle", {}).get("id", "") or vehicle.get("vehicle", {}).get("label", "")).strip()
             _dir = str(vehicle["trip"]["direction_id"])
             try:
                 _line = str(self._data.get("next_departure", {}).get("route_short_name") or "").strip()
                 _dest = self.config_entry.data.get("destination", "").split(": ")[-1].split(" (")[0].split(" - ")[0].strip()
-            except Exception:  # the label is cosmetic, it must never break the update
+            except Exception: 
                 _line, _dest = "", ""
             if _line and _dest:
                 _label = _line + " → " + _dest + " " + (_veh or _crc)


vingerha:
As with other PR, my first question is what you expect as behavior. 
You have to start with the sometimes very (!) limited set of information that the feed provider presents and still provide a unique ID for the map display. There are providers that re-use the trip-id within 48h and there are lines where 5+ vehicles can be on the move on the same route in the same direction.
If I recall this correctly, my main challenges was not the source data but the limited options in the HA-default geojson sensor for having a unique ID. geojson can have more attributes but the current integration is dead wrt to enhancements and not too easy to change.

Pulpyyyy:
Expected behavior: readable marker labels on the map (`40 → Gare d'Orleans 7013` instead of `ORLEANS:Line:40(0)156_bus`) and a stable, bounded set of geo_location entities, sized to the fleet instead of growing forever.

Your three constraints actually argue for `vehicle.id` rather than against it:

- Trip id reuse within 48h: that hurts the current scheme, not this one. The crc32 of the trip_id changes at every run, so each bus comes back under a new unique_id and the registry grows without limit (on my install: 5 entries in half an hour, one already dead). `vehicle.id` is independent of the trip, so identity stays stable across days and the registry stabilizes at fleet size.
- 5+ vehicles on the same route and direction: that is exactly what route_direction_vehicle distinguishes. Each of the 5 buses has its own vehicle id, where 3 digits of crc32 give weaker uniqueness.
- Poor feeds: the fallback is kept. If the feed publishes no `vehicle.id`, the code falls back to the current crc32 behavior, so nothing changes for those providers. On the TAO feed all 94 vehicles carry one.

On geo_json_events: agreed it is limited and frozen, and this PR asks nothing new from it. The unique_id is already built as entry_id + properties.id, the PR only makes properties.id stable and readable. It also cleans up the registry entries left behind by the old crc pattern (only those without a state, at the first realtime refresh).

vingerha:
ok, thanks, this detail is too much for me to review and verify right now. My code is based on the one provider locally and one person afterwards that had issues with the uniqueness, so very little was tested beyond that. The only thing I can remember is that it was not easy to create a unique id, vaguely I also remember that there is a max amount of characters for geojson...t.b.d. later then.

Pulpyyyy:
No rush at all, it can wait until after your holidays.

Leaving a one-glance summary here for later:

- `properties.id`: `ORLEANS:Line:40(0)156` (changes at every run) becomes `40_1_7015` (stable across days)
- `properties.title`: `ORLEANS:Line:40(0)156_bus` becomes `40 → Gare d'Orleans 7013`
- feed without `vehicle.id`: exact current behavior kept (crc32 fallback), nothing changes for those providers

On the max-characters memory: the new ids are 9 characters against 21 today, and the titles stay in the same range, so whatever limit geojson or the entity registry may have, this PR only gains headroom.

vingerha:
OK... at the moment no clue what I did in the pastand why, which will likely reappear when I run some test ...at least I now have feedback from someone that uses it instead of myself :)

vingerha:
So my current view, t.b.d. :)
1. is trying to resolve an issue with geojson, not gtfs2. gtfs2 delivers uniquely id-ed entities in the json file so they can show properly on a map. 
This I will not tackle so easily, I have the same issues with blitzortung where after a thunderstorm I am left with hundreds, sometimes more that a thousand stale thunderstrike-entities. I am using a spook functionality to remove them.
I am hesitant to solve an issue which does not source with gtfs2 so for now and this PR, please remove and make it a separate PR as I need to find out if (?) and how other geojson-using integrations handle this (or not .. like blitzortung)

2. For as far as I recall this is the approach
- As per above, gtfs2 needs to deliver uniquely id-ed entities so they can show properly on a map and have interconnecting lines when the position changes.
- For filtering geojson objects/vehicles on a specific map, the entity as created by geojson needs to have identifiable characters.
- On a single route+direction multiple vehicles can be active. 

This led to an entity with route + direction at the basis and trip_id to make it unique
- route: taken from the gtfs2 start/stop entity
- direction: taken from the rt data as the start/stop entity not always has a direction, this is a real life challenge (see closed issues) and seems to work for now but it remains unsure if the rt data is always providing this attribute
- attribute to make it a unique identifier: I started with the trip_id but then got a issue with another user that had a provider with a very large trip_id so I reduced it with the 'crc' to only 3 characters so a stable trip_id will have a stable crc value.

What I seem to read from above is that you have a similar issue as with the long trip_id but now with route_id which I seem to read being 'orleans:line:40' or?. 
However, the proposed solution assumes that there are semicolons in the route_id attribute which is not the case for most feed providers and also not sufficient if the order is different, e.g. 40:orleans:line returnes 'line' as _line and orleans_line_40 will return the same, i.e. orleans_line_40 as _line and for a TER with route_id 'FR:Line::0c436faa-ccd8-4cb5-9b79-1c8a701364c4:' it will return '', i.e. an empty string

The only thing I can think of is to replace route_id by a self configured attribute or use the name of the entity or something...need to ponder a bit more

Pulpyyyy:
Ok for both. The cleanup is out of this PR, I'll open it separately.

For the route_id, you're right, the split was too fragile. The id is now the raw route_id, the direction and the vehicle id joined with underscores, nothing is derived from the route_id anymore. The title takes the line's short name from the schedule, with the old label as fallback when a feed has no short name or no vehicle id. One more reason for the vehicle id: cut to three digits, the crc only has a thousand values, so two buses on the same line can end up sharing a marker.

vingerha:
The crc is only the same when the trip_id is the same, which can occur within 48h but then the question is ... who needs to separate out the vehicle of a trip of (say) Monday with the one on Wednesday. If I look at the changes, I see hardly any difference with what I provided and am still not sure what it brings.

Pulpyyyy:
Nobody needs Monday and Wednesday apart, that's my point too: the crc separates far more than that, it separates 2pm from 3pm. Every run has its own trip_id, so every run became a new entity, five registry entries in half an hour here, and an event-day line like our 22 mints forty new ones per service day that then sit dead until the next event. The vehicle id keeps one entity per vehicle, which is what a map follows.

The thousand values also run short in practice. Today's TAO schedule alone has five pairs of overlapping runs sharing line, direction and the same three crc digits, and the trip_ids even name the two distinct vehicles:

```
bus 40 dir 0, crc 767: vehicles 4002 + 4007, 07:33-08:19 and 08:06-08:51, 13 min together
tram A dir 0, crc 365: trams A24 + A03, 15:36-15:41, 5 min together
tram A dir 0, crc 634: trams A14 + A17, 16:42-17:15, 33 min together
tram A dir 1, crc 239: trams A08 + A05, 18:22-19:00, 38 min together
tram A dir 1, crc 271: trams A15 + A02, 07:34-08:22, 48 min together, morning peak
```

During each of these windows both vehicles carry the same properties.id, so geo_json_events sees one entity and the marker jumps from one bus to the other. The diff is small on purpose: only the identity source changes, and your crc stays as the fallback for feeds without a vehicle id.


PS: I sent you an invite to the card project. You may also want to give my gtfs2 fork a try (branch integration/all-lots on Pulpyyyy/gtfs2) to see what changed:

- a friendlier config flow: pick a line, directions labelled by their real endpoints (circular lines told apart by where each rotation heads first), stops in order, a suggested sensor name, and the return journey offered in one pass, trains included;
- a two-database import: the full feed lands in a throwaway import base, the live base keeps only the lines you follow, somewhere between 30 and 100 times smaller depending on the network, and sensors never read a half-built base;
- sensors hold to the line and direction you picked instead of matching any trip between the two stops;
- journey durations as attributes, readable entity ids, and the route file with ordered stops written next to the vehicle positions for map cards.

vingerha:
I think I am missing your point still... and as I donot have that much real life examples, only palmbus, what is the difference using vehicle_id or trip_id? 
If 
- 2:10pm with trip_id:A and v_id:001
- 2:20pm with trip_id:B and v_id:002 
- 2:40pm with trip_id:C and v_id:003 

whether using trip_id or vehicle id, it will be three different entities

vingerha:
And a warning to not go too far too quickly. Although I did not cover a lot of options, many of my choices on the current setup have been based on all the feeds that I tested about 50+ If you change the config flow or anything else you need to know all that backlog to be sure that it works in all cases with weird formatted or even missing routes, directions, odd separators, etc. etc.

vingerha:
And I now see you have your own card... nice :)

Pulpyyyy:
In your example the three runs use three different vehicles, so both keys give three entities, agreed. The difference shows when vehicle 001 comes back for the 3:10pm run: with the vehicle id it is still the same entity, with the trip crc it is a fourth. Here a bus loops on its line all day, one vehicle is dozens of trip_ids per day.

On TAO the trip_ids also embed the day's vehicle assignment and departure time, so they change every day and the whole crc set is minted afresh each morning. Palmbus may well behave differently, which is your point about feed variety.

Warning heard. Every change sits behind a fallback: a feed without vehicle id, short name or destination gets the current behaviour byte for byte. If you can name one or two of the weirdest feeds from your 50, I'll run them through this branch before you spend review time on it.

vingerha:
ok, that would work for Palmbus as well I guess, cannot remember. I do remember that I used vehicle id in the past too but no longer sure why I switched this to tripid. So I can give this a try but it may see a revoke

vingerha:
on weird ones, I did not keep a log , need to verify what I have running now.

Pulpyyyy:
I checked the Palmbus feed : all 81 vehicles carry a vehicle.id (plain numbers like 106738), route and direction too, so it holds there as is.

Pulpyyyy:
I went through every vehicle feed mentioned in the repo's issues and probed the ones that need no key, eleven networks, about 20 500 vehicles on the road right now:

| network | vehicles | with vehicle.id | ids look like |
|---|---|---|---|
| TAO Orléans (#173) | 94 | 94 | fleet numbers (7015) |
| Palmbus Cannes | 81 | 81 | numbers (106738) |
| MBTA Boston (#84) | 472 | 472 | y1739 |
| Renfe (#170) | 257 | 257 | 17907 |
| Roma (#126) | 1085 | 1085 | short numbers |
| Bizkaibus (#22) | 213 | 213 | 1101, label empty |
| Adelaide Metro (#65) | 281 | 281 | 1004 |
| Clemson SC (#93) | 58 | 58 | 4162 |
| Zagreb ZET (#146) | 265 | 265 | 442, label empty |
| Pittsburgh PRT (#148) | 247 | 247 | 6350, label empty |
| UK BODS national (#89) | 16871 | 16871 | fleet numbers and plates (BU75WNC) |

None of them derives the id from the trip, and the empty labels cost nothing since the code reads id first. The keyed ones (WMATA, STM, Auckland, Skåne, Vancouver) I could not probe, and that is exactly what the crc fallback keeps covering: it only fires when a feed provides neither vehicle id nor label, and then everything behaves like today, byte for byte.

vingerha:
Thanks, and my comment was meant globally, I actually triggered on your config flow comments.

PR #174, opened 2026-08-19
Export route shape and ordered stops for map cards

Pulpyyyy:
## What

`get_rt_vehicle_positions` already writes the line's realtime vehicle positions to `www/gtfs2/<route>_<direction>.json`, which map cards can consume. But nothing exposes the **planned journey**, so a card can show where the vehicles are, never where they are going.

This PR writes a companion file, `www/gtfs2/<route>_<direction>_route.json`: one `Point` per stop of the journey the sensor follows, ordered by `stop_sequence`. Each point carries:

- `id` (`<route>_<direction>_<stop_sequence>`) and `title` (the stop name), so the geojson integration can consume the file as is;
- `stop_id`, `stop_name`, `stop_sequence`, the scheduled `departure_time` (rendered `HH:MM:SS`, times past 24:00 supported) and the `trip_id`.

The `FeatureCollection` itself carries `trip_id`, `route_id` and `direction_id`.

Points only, no `LineString`: the geojson integration cannot read one, and since the import strips `shapes.txt` a `LineString` could only duplicate the stop coordinates. A map card rebuilds the path by joining the points in `stop_sequence` order.

## File name

Both files share a base name, `<route>_<direction>`, built from ids that come out of the datasource: everything outside letters, digits, dot, dash and underscore becomes a single underscore and the name is lowercased, so `ZOP:653` direction 0 writes `zop_653_0_route.json` next to `zop_653_0.json`. That also covers an id carrying a slash, which used to write into a directory that does not exist and lose the file to an `OSError`. Files written under the previous names stay on disk and have to be removed once by hand.

The point `id` keeps the raw `route_id`: it is an entity id rather than a file name, and changing it would replace every marker in the registry.

## How

- Written from the **static refresh**, so it needs no realtime configuration at all: an entry without any RT url still gets its journey exported.
- Keyed on the drawn trip: the file is only rewritten when the trip changes, not on every refresh.
- Any failure is logged and never breaks the refresh.

## Consumer

The file is read directly (a simple fetch of `/local/gtfs2/<route>_<direction>_route.json`) by a map card I am building; I sent you an invite to its repository. Through the geojson integration the points also work as `geo_location` entities, each named after its stop.

## Verification

This standalone version is compile-checked; its equivalent has been running on a live TAO Orléans install since 2026-08-19 (bus and tram lines, both directions), and the points-only format is rendered by the card on real data.

### DIFF
--- a/custom_components/gtfs2/const.py
+++ b/custom_components/gtfs2/const.py
@@ -1,27 +1,5 @@
 """Constants for the GTFS integration."""
-import re
-
 from homeassistant.const import CONF_OFFSET, STATE_UNKNOWN, Platform
-
-
-_UNSAFE_FILE_PART = re.compile(r"[^a-z0-9._-]+")
-
-
-def safe_file_part(value) -> str:
-    """A route or direction id, made safe to put in a file name.
-
-    Both geojson files are named after ids that come out of the datasource,
-    that is to say out of a url the user pasted: an id like ZOP:653 makes a
-    file no Windows share can read, a percent sign has to be escaped in the
-    /local/ url that serves the file, and an id carrying a slash writes into
-    a directory that does not exist and loses the file to an OSError.
-
-    Rather than list the separators a feed may bring, keep letters, digits,
-    dot, dash and underscore, replace every run of the rest with a single
-    underscore and lowercase, so one route always lands on one file.
-    """
-    return re.sub(r"\.\.+", "_", _UNSAFE_FILE_PART.sub("_", str(value).lower()))
-
 
 DOMAIN = "gtfs2"
 
--- a/custom_components/gtfs2/const.py
+++ b/custom_components/gtfs2/const.py
@@ -1,5 +1,27 @@
 """Constants for the GTFS integration."""
+import re
+
 from homeassistant.const import CONF_OFFSET, STATE_UNKNOWN, Platform
+
+
+_UNSAFE_FILE_PART = re.compile(r"[^a-z0-9._-]+")
+
+
+def safe_file_part(value) -> str:
+    """A route or direction id, made safe to put in a file name.
+
+    Both geojson files are named after ids that come out of the datasource,
+    that is to say out of a url the user pasted: an id like ZOP:653 makes a
+    file no Windows share can read, a percent sign has to be escaped in the
+    /local/ url that serves the file, and an id carrying a slash writes into
+    a directory that does not exist and loses the file to an OSError.
+
+    Rather than list the separators a feed may bring, keep letters, digits,
+    dot, dash and underscore, replace every run of the rest with a single
+    underscore and lowercase, so one route always lands on one file.
+    """
+    return re.sub(r"\.\.+", "_", _UNSAFE_FILE_PART.sub("_", str(value).lower()))
+
 
 DOMAIN = "gtfs2"
 
--- a/custom_components/gtfs2/coordinator.py
+++ b/custom_components/gtfs2/coordinator.py
@@ -28,7 +28,7 @@ from .const import (
     ICON,
     ICONS
 )    
-from .gtfs_helper import get_gtfs, get_next_departure, check_datasource_index, create_trip_geojson, check_extracting, get_local_stops_next_departures
+from .gtfs_helper import get_gtfs, get_next_departure, check_datasource_index, create_trip_geojson, check_extracting, get_local_stops_next_departures, update_route_geojson
 from .gtfs_rt_helper import get_next_services, get_rt_alerts
 
 _LOGGER = logging.getLogger(__name__)
@@ -119,7 +119,18 @@ class GTFSUpdateCoordinator(DataUpdateCoordinator):
                 self._data["gtfs_updated_at"] = dt_util.utcnow().isoformat()
             except Exception as ex:  # pylint: disable=broad-except
                 raise UpdateFailed(f"Error in getting gtfs data: {ex}") from ex
-            _LOGGER.debug("GTFS coordinator data from helper: %s", self._data["next_departure"]) 
+            _LOGGER.debug("GTFS coordinator data from helper: %s", self._data["next_departure"])
+            # The planned side of the map: the journey's ordered stops, written
+            # next to the vehicle positions. It follows the static data, needs
+            # no realtime at all, and only changes with the drawn trip, so it
+            # is keyed on the trip rather than rewritten every refresh.
+            trip_for_export = self._data.get("next_departure", {}).get("trip_id", None)
+            if trip_for_export and trip_for_export != getattr(self, "_route_export_trip", None):
+                try:
+                    await self.hass.async_add_executor_job(update_route_geojson, self)
+                    self._route_export_trip = trip_for_export
+                except Exception as ex:  # pylint: disable=broad-except
+                    _LOGGER.error("Error writing route geojson: %s", ex)
         
         # collect and return rt attributes
         # STILL REQUIRES A SOLUTION IF CONNECTION TIMING OUT
--- a/custom_components/gtfs2/coordinator.py
+++ b/custom_components/gtfs2/coordinator.py
@@ -52,6 +52,9 @@ class GTFSUpdateCoordinator(DataUpdateCoordinator):
         
         self._pygtfs = ""
         self._data: dict[str, str] = {}
+        # the trip whose stops are already exported, so the geojson is
+        # rewritten when the journey changes and not on every refresh
+        self._route_export_trip = None
 
     async def _async_update_data(self) -> dict[str, str]:
         """Get the latest data from GTFS and GTFS relatime, depending refresh interval"""
@@ -125,7 +128,7 @@ class GTFSUpdateCoordinator(DataUpdateCoordinator):
             # no realtime at all, and only changes with the drawn trip, so it
             # is keyed on the trip rather than rewritten every refresh.
             trip_for_export = self._data.get("next_departure", {}).get("trip_id", None)
-            if trip_for_export and trip_for_export != getattr(self, "_route_export_trip", None):
+            if trip_for_export and trip_for_export != self._route_export_trip:
                 try:
                     await self.hass.async_add_executor_job(update_route_geojson, self)
                     self._route_export_trip = trip_for_export
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -24,6 +24,7 @@ from homeassistant.helpers import entity_registry as er
 
 from .const import (
     DEFAULT_PATH_GEOJSON,
+    safe_file_part,
     CONF_API_KEY,
     CONF_API_KEY_LOCATION,
     CONF_API_KEY_NAME,
@@ -991,7 +992,9 @@ def update_route_geojson(self):
         })
     geojson_dir = self.hass.config.path(DEFAULT_PATH_GEOJSON)
     os.makedirs(geojson_dir, exist_ok=True)
-    file = os.path.join(geojson_dir, str(route_id) + "_" + direction + "_route.json")
+    # the ids come out of the datasource, so they are not file names until
+    # they are made ones: see safe_file_part
+    file = os.path.join(geojson_dir, f"{safe_file_part(route_id)}_{safe_file_part(direction)}_route.json")
     _LOGGER.debug("Creating route geojson file: %s", file)
     with open(file, "w") as outfile:
         json.dump({
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -23,6 +23,7 @@ from homeassistant.const import CONF_NAME
 from homeassistant.helpers import entity_registry as er
 
 from .const import (
+    DEFAULT_PATH_GEOJSON,
     CONF_API_KEY,
     CONF_API_KEY_LOCATION,
     CONF_API_KEY_NAME,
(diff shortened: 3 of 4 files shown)


Pulpyyyy:
Field verification done: deployed on a live TAO Orléans install (bus line 40 both directions + tram line A both directions). The four <route>_<direction>_route.json files are written as expected on the first realtime refresh after the trip changes, and a map card renders the shapes and ordered stops correctly. No errors in the logs.

vingerha:
Good idea, thanks :)

vingerha:
I will have a look now, interesting how you were able to test this as my code removes the shapes.txt from the feed zip :)

vingerha:
So I have been playing around a bit, the shapes / LineString still does not work with geojson integration
Can you let me know how you configured the point-based map?

The other issue / improvement is that this has to be outside of the real-time process, else it updates every minute which makes no sense for a route. 
Additional finding with MY data from zou, some trips do not even have a shape , funny. 

My suggestion is to use either shapes OR stops to create a point-based output ... as LineString does not work and Polygon is not applicable....unless you found a solution for representing these.

Then place the async call in the part where the static data gets refreshed in coordinator...that should reduce it already. 

If using shapes then the feed create process needs to be adapted too since it now removes shapes before creating the db.... and with that, unless you see that shapes is better on data, use the stops and not shapes ?

Either way, it would be nice if each point has a identifier that is unique, the naming of the entity does not allow much. 

TLTR; open for discussion :)

Pulpyyyy:
On how I tested it: the TAO zip ships no shapes.txt at all, and your import strips it from the zip anyway, so the LineString in that file has always been drawn through the ordered stops rather than from a real shape. That also covers your Zou finding: shapes never reach the database, whatever the feed carries. Stops are enough for me, and keeping them means the import needs no change at all.

On the point-based map, I do not configure one. My card fetches /local/gtfs2/<file>_route.json directly and draws it itself, it never goes through the geojson integration, which is probably why we are not seeing the same thing. The file already carries one Point per stop next to the LineString, each with stop_id, stop_name, stop_sequence and departure_time, so the unique identifier you asked for is there, stop_sequence being unique within a file. If this file is meant to feed the geojson integration, the LineString should go and only the points stay. If it is meant to be read directly, the LineString earns its place. Which of the two should it be?

On the every minute part, I may be reading it wrong: the call sits in the static branch of the coordinator and only rewrites when the drawn trip changes, so what moves is the trip rather than the refresh cycle. Does that match what you saw?

What I am really unsure about is the key. TAO line 13 direction 0 has 8 distinct paths over 1163 trips:


```
617 trips, 18 stops:  Lycée Hôtelier, Navrin, Charles Rivière, ...
216 trips, 40 stops:  Croix aux Chats, Coin St-Jacques, ... Navrin at stop 10, ...
```
Two arms joining at Navrin. Someone travelling Navrin to Charles Rivière is served by all 8 paths. Keyed by trip, the drawn line jumps between 8 and 46 stops. Keyed by route and direction, the union is 47 stops running through both arms, which no bus does end to end. Taking the most frequent path drops 47% of the trips.

One idea would be a file per distinct stop sequence, written once and pointed at by the sensor, which on TAO comes to 191 files for 86 route and direction pairs. But you know the geojson side and its limits far better than I do.

How do you see it? And do your own feeds show this branching, or is it an Orléans quirk?

vingerha:
1. LineString/Point... : 
please remove the LineString bits as it is unusable, hence my code was never implemented in full....unless you know of a solution via geojson or another map card. geojson supports polygon but this does not apply well for A>B tracks.
Since you are not using the ootb map card to show entities, then this needs to be added to the doc. Is it the custom; map-card? Example yaml?

2. Async call moment:
The coordinator is called every minute to support RT updates.
The code is now in the section that runs when realtime is configured and this section is thus run every minute. Since the geojson only (possibly) changes with the trip_id, it should only run when the static data changes and that is at a different pace, e.g. with me the static reload is every 13 mins for buses and every 18 mins for a certain train as it runs once every hour. Put it higher, outside of the RT section ... only when static refreshes
EDIT: another reason to remove from the RT part is that the path might also be interesting for people without RT data

3. On the key:
I am not 100% sure if I am correct on this so trying to formulate my view.
The map should be showing only the upcoming (!) trip_id-related path, ideally with a RT vehicle location on it. 
For the geojson file using Points, similar as for naming the vehicle-location Points, the geojson entity should have a human readable id and label...e.g. route + direction + ordernumber + uniqueness... the key.
What I believe to be the issue : there are different paths where some skip whole villages or certain stops depending on the (time of) day. When using geojson points then I need to be able to distinguish which points to display, i.e. the ones of the upcoming trip_id. 
I believe this can be resolved with auto-entities and a template that combines the geo_json.XYZ plus the attribute trip_id of the start/stop entity. So the uniqueness should still be defined with the trip_id but not its crc value.... QED
```
type: custom:auto-entities
filter:
  template:
    {{ 'geo_location.b_0_Cannes Gare_' ~ state_attr('sensor.palmbus_b','trip_trip_id')}}
show_empty: true
card:
  type: map
  hours_to_show: 2
```

Corrections and other ideas always welcome :)

Pulpyyyy:
1. LineString is out, the file is points only, and each point now carries an id (route_direction_ordernumber) and a title (the stop name), so the geojson integration can consume it as is. On the consumer: it's a custom card I'm building, I sent you an invite to the repo so you can see it; it fetches the file directly, I'll document the format in the PR text.
2. Moved next to the static refresh, keyed on the drawn trip. Your edit is right and it goes further: without any RT url the line still gets drawn.
3. I'd keep the trip_id out of the id and in the properties instead (it's on the FeatureCollection and I can put it on every point too). The file only ever contains the stops of the upcoming trip, and geo_json_events drops the state of anything gone from the file, so the map already shows just the current path, no auto-entities needed. A trip_id inside the id would mint thirty new entities per trip change, the same registry growth we just chased on the vehicle markers.

vingerha:
For your card it may work without tripid but geojson only imports id and label for as far as I know so in that case I do need the identifier to separate out the trip-path. The only alternative I see is to have the db identify unique paths either via a complex query or a new table generated with feed update

Pulpyyyy:
I checked the geojson integration's code before answering. Its feed manager runs a remove pass on every cycle: any entity whose id is no longer in the file gets a delete signal and removes itself. So with stable ids there is nothing to separate out: the file only ever contains the upcoming trip's stops, and the map through geo_json_events shows exactly those, variants included. When the path changes, dropped stops disappear and the remaining ids move to their new coordinates.

You are right that only id and label cross the integration, but that cuts the other way too: a trip_id inside the id is the only thing your auto-entities template could read, and it would mint thirty new entities per trip change, the same registry growth as the vehicle markers. The trip_id sits in the properties for consumers that read the file directly.

One honest trade: with stable ids a stop entity moves when the variant changes, so a map with hours_to_show would draw its jump as a trail. If that matters, hours_to_show 0 on the stops layer is enough. No new table needed as far as I can see.

vingerha:
For the map card I got it to work without a template, for some reason it does not work with auto-entities template and map (or: I can not get it to work). 

image
image

The one thing that is left is the file-naming, the route 'ZOP:653' will not create a proper file due to the semi-colon. I am not sure what other separators are being used that could create a wrong filename but for now I suggest to replace : and % with a _

vingerha:
And what is this? _route_export_trip  ... I donot see this anywhere created in the code.
And another thought, the user has no choice now, this file is always created with a start/stop setup, not really a problem imo but need a documentation

Pulpyyyy:
Which file for 653 ?
image
image

I do have
```
ZOP:653_0_route.json
ZOP:653_1_route.json
```

vingerha:
image

I guess it is the filesystem to which one is writing

Pulpyyyy:
I seems you are reading files through samba share

vingerha:
Ah, true... windows stuff... on ubuntu it is properly readable but I still suggest to make it less risky and replace : and % plus make all lowercase.

Pulpyyyy:
`_route_export_trip` is the trip whose stops are already written: it gets set right after a successful export and compared on the next refresh, so the file is only rewritten when the journey changes. It was created by its own first assignment and read back through getattr, which is why you could not find it anywhere. It is declared in `__init__` now and read plainly.

On the names I went a bit further than `:` and `%`, rather than guess what the next feed brings: everything outside letters, digits, dot, dash and underscore becomes an underscore, and the whole name is lowercased, so `ZOP:653` direction 0 gives `zop_653_0_route.json`. The positions file gets the same treatment in the same commit, otherwise the two stop sharing a base name. On my own feed that renames all 43 routes, so the files written under the old names stay behind and have to be deleted once by hand.

And yes, the file is always written, one write per trip change. I will put the format in the PR text.

vingerha:
On renaming, yeah that better, I thought of that afterwards but then forgot :)

On const.py (not __init__). I like to keep 'const' with constants only and not have functions, just put them in the helper file for which it is use, after that I am done asking and will merge plus generate a release

Pulpyyyy:
Moved, `const.py` is back to constants only. `safe_file_part` now sits in `gtfs_rt_helper` next to the positions writer, and `gtfs_helper` takes it from the import it already had, so nothing new points the other way. Pushed, and I have put the file format in the PR text.

vingerha:
The entity name needs to be extended with _stop or _busstop else I cannot customize it with an entity_picture via customize_glob (works only on name, not attribs). Any objections?

vingerha:
Also, both vehicles as well as stops have the same elements in the external_id now when being on the same route/direction, which can be helpful with auto-entities as one can have one selection, e.g. with me B_0* shows bus B_0_453 and stop B_0_3. Adding the above extention to the name will help with more detailed filtering too

Pulpyyyy:
No objection, the title only exists for the geojson integration: my card reads `stop_name`, `stop_id` and `stop_sequence` from the properties, so a suffix costs it nothing. I would just put a space in front of it, so the entity_id ends in `_stop` while the name still reads as a name, but either way works for me.

And yes, the shared `<route>_<direction>` prefix across both files was deliberate. Good to hear it lands well with auto-entities.

vingerha:
I now have a setup with the route stops but tbh, as 'points' they have little value imo as it misses the route look&feel alike what your card does. Anyhow, it is an additional feature which at least YOU can use :). When you are ready with your card, let me know so I can add it for testing too and reference it as well in the docs

Pulpyyyy:
On the points: that is the geojson integration rather than the export. It only carries `id` and `label` across, so the stops arrive as isolated markers with nothing to join them. The card fetches the file directly, so it also gets `stop_sequence`, and rebuilds the polyline from it. That is where the route look and feel comes from, and it needs no realtime at all: from the static data alone the line is drawn and its stops are named.

The card is at **1.0.0**, tagged today, and you already have access to the repo. One commit in it matters for you: 0.5.9.8 renamed both geojson files through `safe_file_part`, so the card now applies the same rule when it derives its urls, and `ZOP:653` direction 0 resolves to `zop_653_0_route.json`. Two things on your side then: gtfs2 0.5.9.8 or newer, and the files written under the old names have to be deleted once by hand.

Would you give it a run on your own setup? The README has the install and a minimal config; one sensor is enough, both urls are derived from it. If it holds on your side I will make the repo public, and it is then yours to reference or to test against.

Feedback and opinions very welcome, on the card as much as on what it expects from the integration.

vingerha:
Indeed, and the polyline(string) makes it attractive, not the individual points.
I need to release the latest '_stop' adding element and then make the whole bêta tree into one main release. After that I will be more focused to look at the card with more time :)

Another element I was thinking of is to add RT stamps for each stop on the trip but in the current setup that requires yet another set of attributes to be added to the entity plus not sure if this is more a gimmick then a real use case. 

And yet another thing you may have help/think is something that bothers me since long ... depending on the feed/provider, it is not always easy to build a route based on start/stop. In quite a few case it just cannot find a matching trip based on the selected combination so you need to try-and-try-and-try other combinations, and/or change the direction (as there is none in the feed or all 0), etc.  I also have situations where there are multiple stops with the same name on the same route, beats me why but often only one is working in combi with the other. I thought about grouping them by trip but that is impossible for certain cases, the other solution is to have the end-stop be filtered on the start-stop chosen.




...anyhow...maybe you have other ideas....and please suggest first before spendting time :)

Pulpyyyy:
No rush for the card, whenever your release tree is done.

On the RT stamps per stop: half of it is already in your release. The route geojson from this PR carries a theoretical departure_time on every stop point, and gtfs_rt_helper already decodes the full stop_time_update list, delay and time per stop, it just only picks the requested stop out of it. Both of these are in mainstream today. So the missing piece is only writing those per-stop RT values into the route geojson, a small delta and zero new entity attributes. My card would be a natural first consumer, that would tell us whether it is a gimmick or a real use case.

On the start/stop matching: I hit the same wall and reworked the config flow, on my fork only for now. The direction step no longer shows a raw 0/1 but a label built from the first and last stop of each direction. The stops are then listed in travel order for that direction, same-name stops (platforms, or a circular terminus) numbered in calling order, and a destination placed before its origin is rejected right at the form, so a pair that cannot match any trip never reaches the sensor. One cause is in mainstream as well though: the stop list query merges all trips of a direction ordered by stop_sequence, but partial trips renumber from 0. On my TAO feed, tram B has 110 short runs and the merged list interleaves, real positions come out as 0, 19, 1, 20, 2, 21... That misorder alone makes valid pairs look wrong and wrong pairs look valid, hence the try-and-try. My planned fix, not written yet, is to order by the longest trip of the route and direction; with the order right, your idea of filtering the end stop on the chosen start becomes reliable too. I can show you the flow rework after your release if you want.

vingerha:
not sure if you receive updates but I created two 'issues' / features so we can continue there

PR #175, opened 2026-08-19
Fix issue with rt data when sensors going unavailable after the last departure of the day

Pulpyyyy:
## Symptom

Observed live (TAO Orléans, line 40, realtime enabled, `include_tomorrow` off): the day's last bus departs at 21:17, and at the next coordinator refresh the whole config entry dies —

```
KeyError: 'origin_stop_sequence'
```

The update is marked failed by the `DataUpdateCoordinator`, and **every entity of the entry turns `unavailable` until the next morning** (4 static attributes left, versus ~99 on a healthy sensor). Side effect: the vehicle-positions geojson stops being rewritten, so map consumers keep displaying the last buses of the evening as frozen ghosts all night.

## Cause

- `get_next_departure()` builds yesterday+today's timetable (tomorrow only with `include_tomorrow`) and, after the last departure, finds nothing → returns `{}` (gtfs_helper.py:343).
- With realtime enabled, the coordinator then reads `self._data["next_departure"]["origin_stop_sequence"]` with a **direct access** (coordinator.py:152), while the neighbouring lines use `.get(...)`. `{}` → `KeyError`.
- The exception fires **before** the realtime feed is ever queried, so this is not "no realtime because no bus" — the sensor dies on the way there.

## Fix

Guard the realtime block: when `next_departure` is empty, skip it with a debug log and return the data as-is. The sensors stay **available** with an empty departure list — same behaviour as a realtime-disabled entry after end of service — and recover by themselves at the first departure of the next day.

Status: the failure chain was verified live (logs and entity states above); the one-line guard is compile-checked; deploying it on the same install for an overnight validation is the next step.

### DIFF
--- a/custom_components/gtfs2/coordinator.py
+++ b/custom_components/gtfs2/coordinator.py
@@ -127,11 +127,7 @@ class GTFSUpdateCoordinator(DataUpdateCoordinator):
         if "real_time" in options:
             if options["real_time"]:
                 if not self._data.get("next_departure"):
-                    # after the last departure of the day (without include_tomorrow)
-                    # get_next_departure returns {}: skip the realtime block instead
-                    # of crashing on its fields (KeyError: origin_stop_sequence),
-                    # which marked the update failed and took every entity of this
-                    # config entry to unavailable until the next morning
+                    # when there are no more departures, skip the realtime block
                     _LOGGER.debug("GTFS RT: no next departure for this entry, skipping realtime update")
                     return self._data
                 self._get_next_service = {}
--- a/custom_components/gtfs2/coordinator.py
+++ b/custom_components/gtfs2/coordinator.py
@@ -126,6 +126,14 @@ class GTFSUpdateCoordinator(DataUpdateCoordinator):
         # STILL REQUIRES A SOLUTION IF CONNECTION TIMING OUT
         if "real_time" in options:
             if options["real_time"]:
+                if not self._data.get("next_departure"):
+                    # after the last departure of the day (without include_tomorrow)
+                    # get_next_departure returns {}: skip the realtime block instead
+                    # of crashing on its fields (KeyError: origin_stop_sequence),
+                    # which marked the update failed and took every entity of this
+                    # config entry to unavailable until the next morning
+                    _LOGGER.debug("GTFS RT: no next departure for this entry, skipping realtime update")
+                    return self._data
                 self._get_next_service = {}
                 """Initialize the info object."""
                 self._route_delimiter = None


vingerha:
First, thanks...always good to see people coming with issues and proposals :)
Secondly, I have already had a few AI genreated code thrown at me plus (!) my own AI generated code and most of those PR had issues as AI focuses usually on the particular case without incoporating the overall set of challenges over multiple feed providers and their oddities. 

So, my first question will be (I will repeat in other PR too), what is it that you expect as behavior.
When a entity has no more departures, regardless if it is today or any other day, then what should it show. For me unavailable is somewhat fine. as this is what it is: there is no future scheduled departure today (with/out tomorrow). 

Pulpyyyy:
Thanks for looking at this during your holidays, no rush.

I agree on the state: no more departures means unavailable, and this PR does not change that.

The problem is how the entity gets there. When `get_next_departure()` returns `{}` after the last run of the day, the realtime block raises `KeyError: 'origin_stop_sequence'`. The exception marks the whole coordinator update as failed, so every entity of the config entry goes unavailable and a traceback is logged at every refresh until the first trip of the next morning.

Expected behavior with the patch: same visible state, reached cleanly. The realtime block is skipped with a debug line instead of raising. The guard is not provider specific, it only fires when there is no next departure at all. Users who prefer to see tomorrow's first departure already have `include_tomorrow` for that.

On the AI concern: fair. This was diagnosed live on my install (TAO Orleans, last bus at 21:17, tracebacks from 22:00 on), not by a model scanning the code. The patch runs on that install now, and that is why the PR is a draft: I will mark it ready after it has survived several end-of-service nights, and I will post the results here.

I will be on holidays soon as well, so there is really no urgency to review or merge these PRs. Take your time.

vingerha:
OK, again without having looked at the code, this seems a fix for the rt-data process which throws an error and that I can indeed see as an actual current issue. t.b.d. later ...have a nice holiday too then

vingerha:
I managed to get some sort of connection to the ha-installed code so I can change things locally, using [file shell](https://github.com/junkfix/file-shell). Am in Canada and the connectivity is weird, things are blocked left/right e.g. outlook imap). Anyhow, if you are on discord you can ping me there too, same username :)

Pulpyyyy:
Enjoy your trip in Canada. I'll be taking a break for a week too and put my laptop away. I'll get back to you when I return, likely with some new ideas and logs on the various topics we discussed.

vingerha:
New ideas are great, I had a few in the past but due to loads of feed providers with minimal data, many of them put in th ebin as there was no actual demand.....hence happy with your new feature :) Enjoy the holidays, I took my laptop with me (clearly)

vingerha:
Thanks again

PR #176, opened 2026-08-21
Give SQLite more time to wait when the database is busy

Pulpyyyy:
## The problem

When several GTFS entries start up at the same time, they all open the same
database at once. SQLite allows only one writer at a time, so the others have
to wait their turn.

The connection is opened without a `timeout` value, so SQLite uses its default
of 5 seconds. On a large database that is not enough. The entries give up with
`database is locked`, and because that happens during setup their sensors are
never created at all.

Seen on a 1.1 GB feed with 8 entries: 7 of the 8 failed to set up after a
restart, with errors arriving at almost exactly 5.0 second intervals.

```
06:50:26  Error writing route geojson: database is locked
06:50:51  Error while setting up gtfs2 platform for sensor: database is locked
```

## The change

One line: pass `timeout=60` in the connection string, next to the
`check_same_thread=False` that is already there.

```python
sqlite_file = f"{gtfs_root}.sqlite?check_same_thread=False&timeout=60"
```

Entries now wait their turn instead of giving up. When the database is not
busy, nothing changes: the timeout only applies while waiting for a lock.

## Verification

The SQLite dialect does forward `timeout` from the connection string, the same
way it already does for `check_same_thread` (SQLAlchemy 2.0.52):

```python
>>> d.create_connect_args(make_url("sqlite:///t.sqlite?check_same_thread=False&timeout=60"))
(['t.sqlite'], {'timeout': 60.0, 'check_same_thread': False})

>>> d.create_connect_args(make_url("sqlite:///t.sqlite?check_same_thread=False"))
(['t.sqlite'], {'check_same_thread': False})
```

It is picked up by the dialect's numeric-argument coercion, so it arrives at
the driver as a float, which is what `sqlite3.connect()` expects.

The behaviour was also confirmed end to end: holding an `EXCLUSIVE` lock on a
database for 8 seconds while a second connection tries to write.

```
without timeout (5s default)  -> fails after 7.0s: OperationalError: database is locked
with timeout=60               -> waits, then succeeds
```

That is the same failure the integration hit at startup.

### DIFF
--- a/custom_components/gtfs2/gtfs_helper.py
+++ b/custom_components/gtfs2/gtfs_helper.py
@@ -564,7 +564,7 @@ def get_gtfs(hass, path, data, update=False):
             return
     
     (gtfs_root, _) = os.path.splitext(file)    
-    sqlite_file = f"{gtfs_root}.sqlite?check_same_thread=False"
+    sqlite_file = f"{gtfs_root}.sqlite?check_same_thread=False&timeout=60"
     joined_path = os.path.join(gtfs_dir, sqlite_file)  
 
     gtfs = pygtfs.Schedule(joined_path)


vingerha:
Donot understand this, sqlite can run queries in parallel and a locked state only happens when the db is being updated. Nothing is written to the db only when you update from the feed provider....after that it should be cleanly usable. The code should already take care of it as it checks if an update takes place. So, I need more data or a way how to replicate this. sqlite is pretty fast, I have a 5Gb db for the Netherlands and never had issues so if (!) speed is an issue then maybe I can add some idx.

Re on the test-case: you mention this happens after a restart, a restart of what ? HA? On HAOS or container

Pulpyyyy:
Fair questions, and your point about writes made me look again. Here is what I can document.

**Environment:** Home Assistant Container (Docker), HA Core restart at 06:50. Not HAOS.

**On "nothing is written to the db":** that is what I assumed too, but `check_datasource_index` runs on *every static refresh* of the coordinator, not only on a feed update:

- `coordinator.py:119` → `check_datasource_index(...)` before `get_next_departure`
- `gtfs_helper.py:820-832` → `create index gtfs2_stop_times_trip_id on stop_times(trip_id)` etc.

So on startup, every entry whose refresh interval has elapsed calls it concurrently. The `SELECT count(*) FROM sqlite_master` guard means the `CREATE INDEX` normally does not fire, but the call itself still opens a write-capable connection on a 1.1 GB file, and several entries do it at the same time. That is where my 8 entries collided.

**On speed vs locking:** I agree SQLite is fast and I do not think this is an index problem — the 5 `gtfs2_*` indexes were all present. What I saw was the 5-second default `busy_timeout` expiring: errors arrived at almost exactly 5.0 s intervals (06:50:26, 06:50:51, ...), which is the signature of the timeout rather than of a slow query.

**Reproducing it:** the reliable way is to hold a write lock while a second connection tries to work:

```python
# terminal 1 - hold an EXCLUSIVE lock for 8 s
import sqlite3, time
c = sqlite3.connect("your.sqlite", isolation_level=None)
c.execute("BEGIN EXCLUSIVE"); time.sleep(8); c.execute("COMMIT")

# terminal 2 - meanwhile
from sqlalchemy import create_engine, text
e = create_engine("sqlite:///your.sqlite?check_same_thread=False")
with e.connect() as conn: conn.execute(text("insert into ...")); conn.commit()
# -> OperationalError: database is locked after ~5 s
# with ?check_same_thread=False&timeout=60 -> waits, then succeeds
```

Your 5 GB Netherlands db not hitting this is consistent: it depends on how many entries share one datasource and whether their refresh cycles align at startup. With one or two entries the window is small; I have 8 on one file.

**On the change itself:** it only affects the case where a lock is already held — it makes a waiting connection wait longer instead of giving up. It does not make anything slower when the db is free, and it does not change any query.

That said, if you would rather fix the root cause than the symptom, two alternatives are arguably better and I would be happy to close this in favour of either:

1. not calling `check_datasource_index` on every refresh (it only needs to run after an extraction);
2. `PRAGMA journal_mode=WAL` on the datasource, so readers and the writer stop blocking each other. Note this one needs care: `check_extracting` uses the presence of `<file>.sqlite-journal` as its "still unpacking" sentinel (`gtfs_helper.py:766-773`), and in WAL mode that file never appears.

Happy to test whichever you prefer on my setup.

vingerha:
Kindly stop using claude as a answer/response mechanism as it starts to look like this is me against claude...and then I quit. 

The create index only runs when the idx isn't there, it should have debug statements too if that happens and then yes, if the idx is misisng, it will create it.

vingerha:
Net result, I see you have an issue of database locked but the reasons are invalid so far, so I still cannot identify when exactly this happens except (!) when the db is being refreshed from the feed provider/url/zip. The latter is not easy to being solved as e.g. my NL db takes 2.5 hours to refresh and I will not apply 'wait' mechanisms

Pulpyyyy:
I use Claude to format and to write in English, which is not my first language, but the analysis and the decisions are mine. I will keep it shorter.

On create index you are right, it only fires when the index is missing, and mine were all there.

One point I would keep. The errors came at exactly 5.0 second intervals, and 5 seconds is the default busy_timeout of Python's sqlite3 module. A slow query would give varying times, a bad file would fail instantly. Failing at exactly 5.0 s means a write lock was actually held for several seconds by something, otherwise there is nothing to time out against. So I did not prove what held the lock, but something did.

Given that, and given that a refresh on your NL feed takes 2.5 hours where a 60 s wait would change nothing, I am fine closing this. I will keep the patch locally and come back only if I can catch it happening outside a feed refresh, with proper traces.

vingerha:
Queries can be slow and I can tackle that, the database-locked is unexpected in any of my queries as it should (!) only happen when writing and none of my queries write after the db have been generated. I donot like unexplainable issues hence trying to find a way to reproduce it.

vingerha:
Any chance you had the db open in another tool? Or was the restart after / during the refresh of the db?

vingerha:
Added to that, I quickly searched my code and cannot find a path where the two errors are related too. My new guess (it is a guess) is that the geojson integration is thowing the errors, not gtfs2...cannot prove it but...well

EDIT: nah, forget it, I have geojson active for other things that generate hundreds of entries without any issue.

vingerha:
Can you send a log extract in debug mode with (roughly) 200 rows before and 200 rows after the locked error(s)?

vingerha:
I now discovered where the error comes from, it is from your own PR to add the shape..... I was so confused on how that would pop-up ( _LOGGER.error("Error writing route geojson: %s", ex)). I need to have a look at that one then

vingerha:
I analysed a bit more and still cannot find what would cause the db to lock as this is a write action which happens nowhere, also not in your PR174. It can be a wrong error from sqlite, i.e. wrong wording, it can also be related to a lock from the HA db (which I doubt). Only way to find out is to removed the try/error and let it crash. This I cannot easily do from here with limited access to my server....so after the holidays

Pulpyyyy:
Nothing new since, and I don't have that morning's log anymore. Let's close it, I'll come back if it happens again.

PR #179, opened 2026-08-30
Clean up the stale vehicle markers left in the entity registry

Pulpyyyy:
Split out of #173, as asked there.

geo_json_events registers every vehicle of the positions file as a geo_location entity and drops its state when the vehicle leaves the feed, but the registry entry stays. Since the marker id embeds the trip, each run leaves a new entry behind and the registry keeps growing (on my install, five entries after half an hour, one already dead).

On the first realtime refresh of a sensor with a vehicle position url, this removes the geo_json_events entries of its route that no longer have a state. Only the entries matching the marker id pattern of that route are touched, and a trip that runs again is simply registered afresh, so nothing is lost.

You said you wanted to see how other geojson-based integrations deal with this, so no rush on this one.


vingerha:
I still hesitate, I created another integration in the past for flighttracking via adsb (now disfunct) and that created entities directly in HA so I was sort-of required to remove them. Now it is using geojson and imo it should be done by that. When gtfs2 provides a solution, not sure what else will pop-up or gets deleted by accident. The alternative is to run a orhpane-delete script daily/weekly.... I already do this for blitzortung. 
I now read this for blitz:
Integration can create many new geo_location entities in a short time, especially if the radius is set to a large value. To protect the database from data floods, it is recommended to exclude the geo_location domain from processing by the recorder.
```
recorder:
  exclude:
    domains:
      - geo_location
 ```
I did not notice this before and tbh, I am not sure if I want it set like that ... 
All in all, with your new code and assuming vehicle_id is there in most cases, the orphaned count should drop substantially

EDIT: I just did quick test and geojson does not properly delete the entity (as you wrote too I) and thus it turns 'orphaned', so the orphaned removal script is 'a' way.

Pulpyyyy:
The removal only matches what gtfs2 itself wrote: domain `geo_location`, platform `geo_json_events`, a unique_id ending in the old `<route_id>(<direction>)<crc>` form, and no state at all. Markers created since your merge never match, so this is a one shot migration rather than a permanent behaviour.

Your test matches what I found on the other PR: the state goes when a marker leaves the file, the registry entry stays. So everything minted before the vehicle id change sits there forever unless something removes it, and as it stands my sweep only reaches the routes that still have an entry, which leaves the rest to your orphan script.

If you like, I can narrow it and widen it at the same time: start from the `geo_json_events` entries whose url points into `/local/gtfs2/`, keep only their stateless entities in the old format, and clean those. Since the unique_id starts with the feed's own entry_id the match is exact rather than a guess, so the integration would touch nothing but entities born from files it wrote itself, and it would also catch what an entry deleted since has left behind. Happy to push that, or to close this if you would rather the integration never touch the registry at all.

vingerha:
> ... what gtfs2 wrote ..

gtfs2 does not create geolocation entities, hence reluctant.... a remove-orphaned-script is easiest and uses components that are (!) supported. 

I will keep it open for a few weeks but at this very moment it is a 'no'

Pulpyyyy:
True, the entities are geo_json_events, gtfs2 only writes the file they are built from, so the cleanup does belong outside. I will leave it to you, the vehicle id change already makes the leftover set finite. If it helps, the old markers are the ones whose id ends in (<direction>)<crc>, which is enough for an orphan script or a wiki note.

PR #185, opened 2026-09-01
Fix the rail route_type list, 110 was mistyped as a second 100

Pulpyyyy:
One character fix in the train filter of get_next_departure. The extended rail family reads ...108, 109, 100, 111...: 100 is listed twice and 110 (Replacement Rail Service) is missing. A feed that codes its lines as route_type 110 therefore never matches the train path. Spotted while running my journey tests over several feeds; no behaviour change for any other type.


vingerha:
it is (euh... was) time for glasses...thanks

PR #189, opened 2026-09-02
Clean up the geojson files when an entry is deleted

Pulpyyyy:
Deleting an entry leaves its geojson files in www/gtfs2 for good: nothing removes route_direction.json and its _route.json companion. This adds async_remove_entry to delete them, but only when no other entry still uses the same route and direction (a return sensor for instance).

The names go through safe_file_part exactly like the writers do, and the raw-name files written before the renaming are removed too, so installs that predate it get cleaned as well. The entity registry needs nothing: Home Assistant clears it by itself right after this callback.


vingerha:
Yeah, this is not limited to this one file/config only and I haven't spent much time on it as I had little files in use on production. My initial thought was to just delete all files in www/gtfs2 when GTFS2 initiates, i.e. HA (re)start ... but I never got to that as I used these files in the devevelopment server for multiple/parallel tests, so did not want to loose them with each reboot.
Still this may be the simplest way as all files are always re-created as part of the active entities or realtime processes.

Pulpyyyy:
I consider these two operations to be complementary: this PR prevents the creation of new residual data as soon as an entry is deleted, while a scan at startup cleans up what the history has already left behind. The cleanup appears to pose no risk to active configurations, since everything in www/gtfs2 is recreated (the positions during the next real-time cycle, the route file, and the export note, which is reset upon restart); the only drawback is a blank map for one minute after the restart. Development use cases are the only situation where this should not be done unconditionally. I can add the cleanup here or in a later commit if you’d like; otherwise, this pull request is complete on its own.

vingerha:
If I provide the sweep then it does not need this specific clean-up too. I am just thinking of keeping the code as simple as possible. Assuming that HA users restart HA every (say) month at least, not much can go wrong in file growth as these geojson files are not massive, i.e. a lot smaller than the converted realtime files.

I mean, I do like the setup/feature but it is just more code that only covers a very small part of the overall issue...and it needs testing again.

That said, maybe you can help on ideas on automating testing? I tried to think of that a few times in the past but I got more headaches from all the things it needs to cover for multiple providers and keeping provider data up-to-date.... and then, the result from a testrun will almost never be a fixed figure/time that you can verify as things change over time....here I go again...headache :)

Pulpyyyy:
The problem here isn't so much the quantity as it is the processing involved: if you add a route, retrieve the GeoJSON, and then for whatever reason you delete it, the GeoJSON remains and prevents you from creating an identical new one. Obviously, this happens more often in our tests than in users' actual experiences.

I am also considering implementing a suite of automated tests. But there are two problems: non-regression tests are not relevant in the case of major refactoring. And creating step-by-step tests takes much longer than expected, not to mention that they are not reusable.

vingerha:
Yeah, testing this with all the edge-cases ... again... headache.... 
But.. just as side info, today I had an issue with realtime updates from TER, my dev-side db was 3 weeks old, still provided the correct scheduled departures... but realtime direction changed from 0 to 1.. took me 2 hours to find that out

And this is just the top of the iceberg, loads of gts providers change, make errors, etc. ... so what to test with what data and when? As long as you have no reference (as it is moving) then how to do this?
 My current final thoughts were to create a gtfs2 test data set ... but then I stopped, too much (unpaid) effort

Pulpyyyy:
Side note on your TER hunt: that smells like #193. direction_id is optional in GTFS-RT and protobuf silently returns 0 when a feed leaves the field out, and the SNCF feed is one of those that omit it (that is what prompted the PR). Any change on their side, tagging the field where it was absent before or the reverse, then shows up as a direction flip from 0 to 1 while the schedule stays correct. With #193 an absent field stays absent and the reader falls back to the requesting entity's direction instead of trusting a fake 0. A saved copy of that day's feed would confirm it in minutes.

Pulpyyyy:
On the moving reference: I would not test against live feeds at all. Freeze them once: keep the static zip, and save one raw response of each realtime feed next to it. Replayed offline through the code, the same bytes answer the same question every time, so the reference stops moving because it is a file, not a host. And the day a provider surprises you, that day's capture joins the set as a regression case: your TER hunt is the perfect example, a saved copy of that feed turns two hours into minutes and the case stays reproducible forever.

For what to assert, I would not compare exact outputs, they die at every refactor. The promises the sensors make age much better: a departure carries a stop and a time, the realtime direction agrees with the static side, a service crossing midnight still ends after it. Those are about the data contract, not the implementation, so they survive rewrites, and most of them can be generated from the fixture itself instead of being written by hand.

The when then costs nothing: offline bytes make the run free, so it can happen at every change rather than on a schedule.

vingerha:
Thx... ideally the test would provide its own test data in the script, not files to read from but it is exactly that query element that differs. The other challenge is that there is a lot of comparing with 'now' so datasets need to be up-to-date too. What could be done is to create a datasets that simulate the output of a query, for each case a differnt set and the dataset could be 'updated' to today's date via a script. ... just thinking out loudly, motivation is low tbh.

Pulpyyyy:
The two things you raised are what the tree ended up doing, so neither costs anything now.


The data does provide itself: there is no expected file anywhere. For each route in a fixture's manifest, the distinct stop sequences its trips ride are read from stop_times, and that is the expected side. The feed states what the line rides, the script only asks the question.


And the 'now' never has to be moved. The clock is pinned to a day those trips actually run, in the agency's own zone, so nothing ages: no dataset to bring up to today, and no script to maintain for it.


That is 95 cases over the four feeds as it stands, 59 passing and 36 carrying a known-failure mark with its reason.


Which leaves this PR needing a yes or a no rather than more discussion. If you would rather have the startup sweep alone, say so and I will close this one. The difference between the two is only the window: until the next restart, a leftover file blocks recreating the same route.

PR #190, opened 2026-09-02
Stop doubling the sensor name in the entity id

Pulpyyyy:
The departure sensor names both its device ("GTFS - name") and itself ("name"), and Home Assistant composes the two, so every new sensor lands as sensor.gtfs_x_y_x_y with the name twice. The entity now declares has_entity_name and no name of its own, so it takes the device name: sensor.gtfs_k8_orleans_paris_austerlitz for example, and the two directions of a line come out consistent with each other.

Entities already registered keep the id and name they have; only newly created sensors change.


vingerha:
Thanks, it is however not 1-2-3 to get rid of the old ones if you use the same name in the config....it returns from .storage entities deleted items...so again with the same double name. Only option is to change the name a bit, e.g. add a - or something :)

PR #191, opened 2026-09-02
Dispose the previous schedule before opening the new one

Pulpyyyy:
Small follow-up to this morning's 4883da5. Moving get_gtfs to the top of the local stop coordinator refresh left the dispose block halfway down the method, where it now closes the engine that was just created, while the previous cycle's engine is only released whenever the garbage collector gets to it. Since get_gtfs builds a new Schedule on every call, moving the dispose above get_gtfs gives it back its original target, the previous cycle's schedule, and the db file handles are released at a known point again.

No behaviour change otherwise: disposing the fresh engine was harmless (connections reopen on use), this just restores the explicit release order the code had before.


vingerha:
thanks, did not notice that it had shifted...copy/pasting secions bares a clear risk

PR #192, opened 2026-09-02
Put the aerial lift and funicular icons back the right way round

Pulpyyyy:
Two extended route types map to the wrong icon: 1300 (aerial lift) showed an airplane and 1400 (funicular) showed the gondola. 1300 now gets mdi:gondola and 1400 mdi:stairs, the closest MDI offers for a funicular.


vingerha:
thx :)

PR #193, opened 2026-09-02
Only convert direction_id when the realtime feed actually sets it

Pulpyyyy:
direction_id is optional in the GTFS-RT spec and protobuf hands back the default value 0 when a feed leaves it out (the SNCF and TTC feeds do). convert_gtfs_realtime_to_json wrote that 0 into every converted trip, so every untagged trip came out labeled direction 0 and the reader's fallback for an absent direction ("nn") was unreachable. The key is now written only when the field is explicitly set; the reader already handles its absence. The SIRI path is untouched.

### DIFF
--- a/custom_components/gtfs2/gtfs_rt_helper.py
+++ b/custom_components/gtfs2/gtfs_rt_helper.py
@@ -577,11 +577,15 @@ def convert_gtfs_realtime_to_json(gtfs_realtime_data):
                     "start_time": entity.trip_update.trip.start_time,
                     "start_date": entity.trip_update.trip.start_date,
                     "route_id": entity.trip_update.trip.route_id,
-                    "direction_id": str(entity.trip_update.trip.direction_id)
                 },
                 "stop_time_update": []
             }
         }
+        # direction_id is optional and protobuf returns 0 when a feed omits
+        # it, which reads as a genuine direction and mislabels every untagged
+        # trip; leave the key out instead so the reader falls back to "nn"
+        if entity.trip_update.trip.HasField("direction_id"):
+            entity_dict["trip_update"]["trip"]["direction_id"] = str(entity.trip_update.trip.direction_id)
         for stop_time_update in entity.trip_update.stop_time_update:
             stop_time_update_dict = {
                 "stop_sequence": stop_time_update.stop_sequence,


vingerha:
I am going to reject this, not because I am sure it is not going to work, because I am sure that the current works, this code is not nice but based on a lot of issues with direction id, not just being 0 but being missing from static, or always 0, missing from static and rt and always 0 on rt or even combinations... I can not (do not want to) retest this as each code change was done on a per case level. Claude has no clue about the case background.

So unless you have a bug, code improvements may not pass, direction_id sucks across the globe

Pulpyyyy:
@vingerha
 Could you check the TER bug you ran into? I suspect it's related to this change.

One thing to keep in mind is that Claude doesn't automatically verify its output with real-world tests. I found the SNCF issue during testing and implemented the fix.

Not seeing a bug right away isn't necessarily proof that the code is behaving correctly.

vingerha:
The TER bug was related to old data, the old db contained a 0 in the db for that specific trip and the new one a 1, not anything to do with the code. I will keep this open for a while but I really hesitate to change running code to something that I cannot properly test. The direcoitn_id was the biggest nightmare between both start/stop and local-stop setup. Not saying all feed providers are poor but there are really poor datasets out there, some even so poor that they can not be used in gtfs2 (Mallorca)... and others donot prio the data at all...my zou realtime has beein incorrect for about 1,5 months, they changed the trip/route id in the staic one and then again and never aligned with the rt...close to 2 months now without rt....sigh (latter has nothing to do with direction, just needed to complain )

Pulpyyyy:
I don't have you test data/history, but i  think the key point to verify is whether the 0 actually comes from the GTFS-RT feed or from protobuf's default value handling.

If that's the case, then the current code is effectively treating "direction unknown" as "direction = 0". Those are not equivalent states.

I'm not claiming the current behaviour always produces visible bugs, but it does mean we're potentially injecting a value that wasn't present in the source data. A real direction_id = 0 should obviously be preserved, but an absent field should remain absent.

So for me the real question is:

- Is hasDirectionId() true and the feed explicitly sends 0?
- Or is hasDirectionId() false and we're only seeing 0 because it's protobuf's default value?

If it's the latter, then we're not reading source data anymore, we're making an assumption about it.

vingerha:
Yes, agree that it is possibly providing an incorrect situation as it could in reality be either 0 or (!) 1... but it is intentionally set to 0 so I can treat the data properly later in the code/selection. ...again, it works until a bug applies.

Pulpyyyy:
By converting an absent field to 0, we're no longer preserving the source data. We're introducing an assumption that may or may not be correct.

IMO that can generate false test results, misleading debugging information, and inaccurate data analysis because there's no longer any way to distinguish

I'll have a look at ZOU +RT with my fork, didn't notice problems but didn't spend many times with zou

vingerha:
I agree it could (!) be an issue but until it is proven ... not willing to take action.

On zou, this is a known case, no need to spend time https://transport.data.gouv.fr/resources/82182

vingerha:
I will try to label this as draft or something for 'later'. Maybe this is a good case to create the first automated testscripts for, after which this PR may pass too.... I hope claude can help with that as (test) data creation/parsing etc. is a pain in the ...

Pulpyyyy:
Sure Claude does ;)

vingerha:
Do you have time for a creating an inital test setup? Seeing the load of PR's you created, I would like to be better supported with testing as the code can still use a lot of polishing but the manual review is substantial as (repeating) the code has grown case based...so if I can add the case to a testcase then the next time I just need a pyttest to run, instead of poking my poor old braincells.

My idea is to use multiple datasets static assuming the db / query works and multiple datasets realtime to verify combinations. The challenge is to start 'small' ...so onlystatic first. Later the sql elements can be added but that is a different beast 

If you donot have time then I will start with it,

Pulpyyyy:
Will have less time to spend with GTFS but will have next week

vingerha:
wil ltry to give it a start then to see if my ideas are feasible. The datasets need to be easily adaptable for new cases and remain modest in size ... QED

Pulpyyyy:
I already have several tools that I can share. However, they are still fairly raw and need some cleanup and organization.

As for the dataset, I can easily provide a subset of the GTFS data (similar in size and processing speed to the GTFS extracts I previously provided). Since we already have a list of providers, it should be straightforward to prepare and share the relevant data.

vingerha:
I am already working on it. First issue is that the sql is too much embedded in e.g. get_next_departure so for testing I still need a db and this I donot want as the db updating is a nightmare just-for-testing. I will split it out first, which will also allow separate db-testing in the future

Pulpyyyy:
GitHub CI can handle create and initialize the database on the fly
But you will need to push code before testing, bad side effect

vingerha:
I just finished a first crude test setup, claude can really be a pain ... anyhow, biggest issue is that the test requires homeassistant to be installed (as gtfs_helper.py uses stuff), for just two simple procedures, I will look at this to see if it can be circumvented.

vingerha:
I now added a synthetic test, reduced to a few trips, this cen be altered by hand to copy/amend for new cases.
maybe tomorrow I wil start in a case with realtime data, the effort is in getting a working set of data, extract the various in/output and then make it synthetic. 
From my **current** pov, there would need to be about 10 cases per rout or local_stop. 
For the db, I now doubt if this is needed as this is in active use so under daily live testing. 
I donot plan to test the config_flow either, I think it is too complex to have the test simulate entity creation to then verify if the naming / attr. are ok.
Then also want to test file creation, from my pov last in the prio.

Pulpyyyy:
For workflow i do have a renderer (outside HA) so i't not real test but give a WYSIWYG

vingerha:
Let's see, it does not have to be perfect, I want to cover those elements with th biggest risks and with the most annoying testing activities. Als tests need to be maintained with changing code

vingerha:
But the best thing that all your effort brings, is that I can start on improving the code in the near future, admitting the direction_id is on that list. There were times that I did not spend time on it for 6months plus and a lot of reasoning behind the code is forgotten, likely still valid but no longer 100% sure so it became a bug-only effort

Pulpyyyy:
Test material, as promised. Three small PRs, in this order:

1. #203: `tests/ha_stub.py` plus a conftest, so `pytest tests/` runs without installing Home Assistant. It loads gtfs_helper under a synthetic package, the real `__init__` never runs, and the clock stays pinnable. Your test_static_cases.py does not change. It is the #200 file again in your layout; tell me the typo you spotted so it does not come back.
2. #204: Fixtures: real feeds cut down to a few routes, 24 to 88 KB each (Palm Bus, TAO, SNCF with its trip updates and alerts, GVB), with a manifest saying where every byte came from, what was kept and why, plus the two scripts that build one from any zip or RT url. New cases stay small and reproducible.
3. #205: A journeys suite on those fixtures, as pytest: for every route and direction it checks the stop list, a valid origin/destination pair and the swapped one, through the real query on a db built from the zip, no pygtfs, about 12 s for the five fixtures. It asserts promises rather than exact output, so it survives refactors. On your main it reports 69 failures today, all known: the stop selector (#198), the swapped pair answering instead of empty, trains matched on another line. Each comes with its fix, so the PR and the test that turns green arrive together.

The stub needs the standard library only; the rest needs nothing beyond what your requirements.txt already lists.

vingerha:
So, I made a bit of progress too, admittedly slower then I initially thought but that is because of my control(freak) approach, With your HA stub and kicking Claude a lot, I was able to provide a minimal install in GH so the tests can run after each commit/pr/etc.  Since your tests cover a much larger scope you likely need actual installations like pytgfs / sqlalchemy /protobuf / etc. ... on GH. Have you verified that this installs properly (and: fast enough)? Can you supply your cases as requested in a separate tree or has this already been done? I am a bit overwhelmed with the mass of stuff you supply so natural instinct is to put the brakes on.

Pulpyyyy:
Measured here, not on GitHub: the seven packages install in 51 s from an empty venv, sqlalchemy and protobuf come as wheels. The separate tree is #206, replacing #204 and #205; #203 is closed, your ha_stub covers it. After that I stop pushing unless you ask.

vingerha:
Back to the original case of this PR, do you have a real life feed where the direction_id is missing?

Pulpyyyy:
Yes. The SNCF trip updates feed on transport.data.gouv.fr never sets it: 0 of 9 in the frozen fixture (26 Aug), 0 of 1106 in the live feed this morning, checked with HasField, not on the value. The static side of the same feed does carry directions, 105 of the 182 fixture trips ride direction 1, and after conversion on main all of them come out as 0.

To be fair to your "it works": on SNCF it does. route_id is empty in that feed too, so the reader switches to trip mode and overwrites the direction with the sensor's own before matching. The 0 never gets compared. It bites on a feed that sets route_id and leaves direction_id out, in route mode: a direction 1 sensor never matches, a direction 0 sensor takes both ways. TTC is one I noted as such, but I have no capture of it, so I am not asking you to take that on my word.

Proposal: I capture such a feed into the fixtures and add the route-mode scenario, failing on main and passing with this change. The PR stays draft until it has that case.

vingerha:
I now see the coding issue, or a for-me unexpected behavoir, I woul dhave expected that string() to return a empty string "" if not available but it returns "0"... python sometimes does surprising things....at least, I am surpised

Pulpyyyy:
Not Python this time, protobuf. GTFS-realtime is proto2, where `direction_id` is an optional uint32, and reading an unset optional scalar gives you the field's default rather than nothing: 0 for a number, "" for a string. So `str()` never saw an absence, it was handed a real 0 and faithfully wrote "0".


That is also why it stayed invisible for so long. `route_id` is a string, so a feed that leaves it out reads as "" and looks absent, exactly as you would expect, while the same feed leaving out `direction_id` reads as 0 and looks like a genuine direction 0.


Presence is kept apart from the value, in a bit only `HasField` can read. It is the only way to tell "the feed said direction 0" from "the feed said nothing", which is what the change asks before writing the key.

vingerha:
I see but then this fix only covers that single field and that may not cover all missing attribute cases.
The idea behind my setup is to first convert all to json and use that json in the rt process. 
This allows to use a .rt file or a .json file as input
As the convert-to-json did not use hasfield for any field, having it only for direction_id misses other fields that may not be provided by the .rt protobuf file and it sets incorrect values too.

Pulpyyyy:
You are right that it covers one field, and the converter does write every other one unconditionally.


The asymmetry is only on the numbers though. A string a feed leaves out arrives as "", which is falsy and reads as absent, and the reader already leans on that: at the stop match it tests `stop_id == ""` explicitly and falls back to the sequence. So trip_id, route_id, start_time, start_date and stop_id are safe, if a little by accident.


Where the default cannot be told from a real value is the delays and the sequence.


`arrival.delay` and `departure.delay`: 0 means on time. The picker takes the greater of the two, which quietly rescues a side that is not set as long as delays are positive, an absent departure delay of 0 losing to a real arrival delay of 300. It only misfires on a vehicle running early, where the spec allows a negative delay and 0 wins over -60. I have no capture of a feed publishing one, so that part is on the spec rather than on something I have seen.


What the frozen SNCF capture does show is the absence itself: of 240 arrival and departure sides, 20 are not sent at all while 123 carry an explicit 0. The feed separates a side it states as on time from a side it does not send, and the conversion reads both as 0.


`stop_sequence`: 0 is a valid first stop, and it is exactly the value the reader falls back on when stop_id is empty.


`arrival.time` and `departure.time` are the same family but they land safely, the code takes the max of the two and treats 0 as no value, so a side that is not set never wins.


So yes, direction_id is the one that bit me and the only one I fixed. The delays and the sequence want the same treatment, and a sub-message that is not sent is detectable the same way.

PR #194, opened 2026-09-02
Give a local stop sensor only its own stop's departures

Pulpyyyy:
A local stop sensor picks its rows out of the coordinator data with a startswith on the sensor name. The name begins with the stop_id, so a sensor whose stop_id has another stop's id as a prefix also claimed that stop's departures (a sensor for stop 10001 matched stop 1000 too, and the last match won). It now compares the stop_id itself.

### DIFF
--- a/custom_components/gtfs2/sensor.py
+++ b/custom_components/gtfs2/sensor.py
@@ -556,7 +556,10 @@ class GTFSLocalStopSensor(CoordinatorEntity, SensorEntity):
         self._attributes["next_departures_lines"] = {}
         if self._departure:
             for stop in self._departure:
-                if self._name.startswith(stop["stop_id"]):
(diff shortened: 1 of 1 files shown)


vingerha:
Need to ponder over this a bit as in a first throw the name alone seems to make no sense but I did this with a reason. At least one use case could be a train station with multiple platforms where I would not want 9 different entries refering to stop_id that donot indicate the platform itself ... This needs carefull testing and maybe a tweak....

vingerha:
I re-read your comment and understand the issue with 1000 vs 100 ... Then I looked at the commit form 2024 where I changed it to contains from a hard compare.... and that was due to another change in that commit, The erros: I just continued to filter using the name rather than the _stop which was already known...so good to go :)...thx

PR #195, opened 2026-09-02
Fix the NameError at the end of get_trip_stops

Pulpyyyy:
get_trip_stops opens its db handle under the name schedule, but the line before the return disposes _pygtfs, a name that only exists in get_route_departures, so every extract_trip_stops service call raises NameError just before returning its result.


vingerha:
Thanks, this is a very old thing and result from a copy of the start/stop. I always wonder why things like this never lead to issues. I mean, I know it does not do a thing in its current form ... but then why have a 'dispose' in there in the first place

PR #196, opened 2026-09-02
Append the query-string api key only to the urls that are set

Pulpyyyy:
The three realtime urls default to None, and with an api key in query_string the concatenation ran on all three, so an entry that leaves vehicle_position_url or alerts_url empty crashed its whole refresh with a TypeError (None + str). The key is now appended only to the urls that are actually set, in both coordinators.


vingerha:
Thx...related to this, if you have time... I noticed that when I add a verhicle url to the start/stop, I could not remove it afterwards so had to delete and recreate the whole entity....if you experience this too then please create an issue, may also apply to the alerts

PR #200, opened 2026-09-03
Run the test suite without installing Home Assistant

Pulpyyyy:
This is one file that lets tests/ run without installing Home Assistant.

tests/ha_stub.py stands in for the homeassistant modules the gtfs2 import chain touches, and loads gtfs_helper on its own, so `__init__.py`, and the coordinator and the platforms behind it, never run. `_interpret_departure_rows` only ever asks hass for `config.time_zone`, so there is nothing a real HA has to answer for in these cases.

Why I think it is worth it: installing Home Assistant is a very heavy dependency for a suite that reads dicts and datetimes, and it ties that suite to an HA version. Without it, installing and running takes seconds.

I did not touch the test itself, only the import at the top. freezegun still drives the clock, since the stub reads `datetime.datetime.now`, which is what `freeze_time` replaces.

If homeassistant is installed anyway, `install()` does nothing and the test reads the real `dt_util`. const, gtfs_helper and gtfs_rt_helper load under the stub. coordinator, sensor and config_flow do not, and I would not want them to: what a test proves there is that the integration speaks to HA correctly, which is exactly what a stub cannot answer for. It raises with the name of what is missing instead of pretending.

Ran here with no homeassistant installed, case_1 passes.

### DIFF
--- a/tests/ha_stub.py
+++ b/tests/ha_stub.py
@@ -1,0 +1,343 @@
+"""Import one gtfs2 module without a Home Assistant install.
+
+gtfs_helper imports homeassistant at module level, and importing it the plain
+way (`from custom_components.gtfs2.gtfs_helper import ...`) also runs the
+package `__init__.py`, which pulls in the coordinator, the platforms and with
+them a good half of Home Assistant. Installing homeassistant answers all of
+that, but it is a heavy pin for a suite whose subject is the integration's own
+logic: `_interpret_departure_rows` reads a list of dicts and a few datetimes,
+and never touches a hass object beyond `hass.config.time_zone`.
+
+So the handful of homeassistant modules the import chain touches are registered
+as stubs, and `load()` imports the wanted module on its own, under a synthetic
+package whose `__path__` is the component directory, so its relative imports
+(`from .const import ...`) still resolve while `__init__.py` never runs.
+
+Only the functions the code actually calls carry behaviour: utcnow, now, as_utc,
+utc_from_timestamp, get_time_zone and parse_datetime. Everything else is an
+empty shell, and a shell that is reached raises instead of returning something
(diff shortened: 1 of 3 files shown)


vingerha:
LOL... I spent quite a bit of time with Claude on this as not actually my 'thing' and Claude concluded that it woul dnot work ...sigh

vingerha:
My 'claude' using Sonet 5 (medium)
```
You're right, and I was wrong. Let me pull the actual diff to see precisely what technique they used, since the description reveals something I missed.

You're right, and I was wrong to say this "becomes a nightmare" — this PR solves exactly the problem I hit, using a technique I didn't consider: loading gtfs_helper.py directly via importlib, under a synthetic parent package, so the real __init__.py (which is what dragged in coordinator/sensor) never executes at all.
```

vingerha:
And this is one of the downsides of AI, when one is not up to speed in a specific area then AI can push you anywhere

Pulpyyyy:
For what it is worth, I see Claude as a good complementary tool: it goes through code faster than I do. But the time spent framing the context and the need properly is always won back at the end. Asked cold it answers from the general case, and the general case here is that you cannot import a HA integration without HA. Given the actual import chain, it finds the way through.

The two parts that stay on me are the context going in and the read coming out: saying what the code really is, and going over what it gives back before it goes anywhere.

And the moral of your quote is that it will happily tell you "You're right, and I was wrong" whichever side it was on five minutes earlier. Being agreed with by an AI proves nothing, which is why the human still has to be the one who is right.

vingerha:
Indeed, after a session with Claude I am always 'right' ...whcih is what I tell my kids all the time :)

vingerha:
I copied the code so closing this without commit (there was one typo)

PR #201, opened 2026-09-03
Fill the keys the translations were missing

Pulpyyyy:
The four translated languages had drifted from `strings.json`: 21 keys missing in French, 19 in German, 19 in Spanish, 5 in Portuguese. German and Portuguese also still carried `stop_limit_reached`, which strings no longer declares. Home Assistant falls back to English for a missing key, so those fields showed English inside an otherwise translated flow.

Most of the gap is the `update_gtfs_rt_local` service fields and the `api_key_location` selector, which had never been translated outside Portuguese, plus `source_timezone_correction` and `check_source_dates` on the options screen.

I translated your English, and followed the wording each language already uses rather than imposing my own. Portuguese was the model for the service fields, being the only language that had them. No code, no behaviour change, no existing string touched: additions only, plus the two stale keys removed. Nothing is reformatted, so the diff is only the new lines.

If it helps, I am happy to keep the five files in step whenever a screen changes, rather than leaving them to drift again.


vingerha:
thanks, I in deed dono tmaintain non-english stuff and for a next time, string.json is not the leading bit, when translations apply it goes to the en.json by default.

PR #203, opened 2026-09-04
Run the test suite without installing Home Assistant

Pulpyyyy:
#200 again, in the layout tests/ has now. `tests/ha_stub.py` stands in for the homeassistant modules gtfs_helper imports, and claims the `custom_components.gtfs2` package name so the plain `from custom_components.gtfs2.gtfs_helper import ...` loads that module on its own: `__init__.py` and the platforms behind it never run. conftest.py does both before any test module is imported, so test_static_cases.py does not change. Where a real Home Assistant is installed the stub steps aside and the real modules are used.

Verified on a machine without homeassistant: `pytest tests/test_static_cases.py --tb=short` passes case_1 and case_2, results.txt comes out identical. requirements.txt drops homeassistant and adds tzdata on Windows only, where Python has no zone database of its own.

An import the stub does not carry says so by name instead of reading like a broken install. coordinator, sensor and config_flow stay outside on purpose, a stub cannot answer for them honestly.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,10 +7,24 @@ re-running pytest or scrolling terminal history.
 
 Only concerns itself with `test_static_case` (test_static_cases.py);
 other tests in this tree are unaffected and not written to the report.
+
+Also registers the Home Assistant stand-ins of ha_stub before any test
+module is imported, so the suite runs without a homeassistant install.
 """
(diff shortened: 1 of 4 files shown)


Pulpyyyy:
Closing, your ha_stub covers it; the provider tree is #206.

PR #204, opened 2026-09-04
Add frozen feed fixtures and the two scripts that build them

Pulpyyyy:
Second of the three. Five real feeds cut down to a few routes, 24 to 88 KB each, under tests/fixtures: GVB trams 1/7/13/14/17 (a third of the trips carry the wrong direction_id, one circular line), Palm Bus 21/22/A/B (circular, branch variants), SNCF K8+/P8 (trains matched by stop name, a line that exists three times in the feed), TAO 22/N/40/41/A/B (circular, night line, weekday only, interleaved short runs), and an SNCF capture with its trip updates and service alerts plus scenarios.json, the journeys those entities make testable. Each manifest.json says where the bytes came from, what was kept and why.

build_fixture.py cuts any static zip down to named routes, a few trips per distinct stop pattern, so every shape a line runs survives with a few specimens. It reproduces the four static fixtures byte for byte from their source zips. capture_rt.py freezes a static + trip updates + alerts capture: SNCF by default, any feed through --static/--trip-updates/--alerts, and --source-dir re-reads files already on disk, a fixture directory included, which is also the way in for a feed that wants an api key.

On the moving reference: a test pins its clock at the manifest's captured_at (freezegun or ha_stub.freeze), so the data never has to be shifted to today. Nothing reads these yet, the journeys suite comes next and is what they are cut for.


vingerha:
You are going too fast, I am trying to create a non-feed specific set of tests as I also mentioned that tests need to be maintained. When you add feed-specific tests then I have no clue what it does and why. 
In general I do not intende to maintain feed-specific tests, so the risk is that if they fail at some point in the future, I will just delete them.
Please slow down, provide ideas first

Pulpyyyy:
I'm only sharing tools and data I already had, not trying to impose a large set of feed-specific tests.
The value I see is in the real-world edge cases they contain, which could help define feed-agnostic test scenarios.

Pulpyyyy:
Replaced by #206, in tests_provider/ as asked.

PR #205, opened 2026-09-04
Check every journey the fixtures offer through the real query

Pulpyyyy:
Third of the three. For each route named in a fixture's manifest and each direction it runs, one test per promise: the stop list offers each stop once, in an order every trip agrees with, and folds two platforms of a place called one after the other; a valid origin/destination pair gets a departure on the right stops, in riding order, arriving no earlier than it departs; the swapped pair gets nothing, or a ride the line really makes. Trains are checked by stop name and must stay on the asked line. The checks call get_stop_list and get_next_departure as they are, on a db fixture_db.py builds from the zip in a second and reshapes the way pygtfs stores it, with the clock pinned by freezegun to a day the trips run, in the agency's zone. They are promises about what the sensors say, not about how the query says it, so they survive a rewrite of the query.

95 cases, 17 s here. 36 are known to fail on main today and are marked xfail with the reason, strict: the stop selector (#198), the swapped pair answering a departure the line does not ride, and trains matched on another line. The fix that lifts one has to drop its mark, which is how a fix and the test that turns green arrive together. Stacked on #203 and #204, the diff shrinks to this commit once they are in.


vingerha:
same as for my other comment, I already mentioned that I am not in favor of testing db, same risk: if it fails, I will delete them.
It needs to be simple and mangeable and only cover main risks first....

Pulpyyyy:
From my perspective, #205 is actually an attempt to validate generic behaviour across multiple providers rather than introduce provider-specific tests. The fixtures are just real-world inputs.

The test logic itself is provider-agnostic: the same assertions are applied regardless of where the data comes from. What changes is only the input dataset and the edge cases it exposes.

I'll admit I'm a bit disappointed that it is being viewed as provider-specific testing, because that was never the intent.

In any case, that's fine. I'll keep this work in my fork. If you ever find parts of it useful, whether the fixtures, tooling, or test ideas, feel free to pick and choose whatever makes sense for the project.

vingerha:
There are two things at play here:
1. understand what is being tested.
From this mass amount of files which deviate to what I am working on, the problem is that I cannot easily identify the case. It does not follow my current test-logic so it is quite a challenge to see what is being tested in the first place. If in the future it fails then what to look at that needs to be repaired, the data(source) or the code.... the current code contains lots of if/then statements to deal with incomplete data so I need scenario to see if that all works when code changes

2. maintenance
I am not negging the value of real feed provider tests but if it is not maintained then a feed-db in the future cannot be guaranteed to being covered by tests of the feed-db today, e.g. zou changed their route_id and trip_id to contain ':' and this crashed the code. If I would use tests for the zou database of the before-situation then I would never detect this until a bug is repoted. Then updating a testcase with new db data is a lot of work

Hence I aim for pure synthetic cases which have been pre-designed on coverage, which is what I am doing now...i.e. designing what I really want to test and what cases I cannot care less.

In the end, I am happy and OK to add your tests and I see their value but they should not overwrite my testing setup because that is the only one that I will maintain. So please put them in a folder called /tests_provider or something similar at the same level as /tests

Pulpyyyy:
I think there is another aspect to maintenance that may be getting overlooked.

A failing provider-based test is not necessarily something that must be fixed. Sometimes it is simply a warning that something changed, and that information has value in itself.

If a test that used to pass starts failing, the change is recorded. We can then decide whether it is a regression, an expected change in the source data, or a behaviour we no longer care about.

Taking the ZOU ":" example, having fixtures from before and after the change would not necessarily have prevented the breakage, but it would have made it immediately visible when the assumption stopped being true and exactly which datasets were affected.

More generally, I see synthetic tests and provider fixtures as solving different problems. Synthetic tests provide targeted coverage for the scenarios we intentionally design. Provider fixtures accumulate real-world scenarios and edge cases that we did not necessarily anticipate.

Also, I don't have the historical knowledge of the project, nor access to all the problematic cases that have been reported by users over the years. The only thing I can realistically contribute is generic testing across a variety of datasets to surface behavioural differences.

Those tests are not meant to decide whether a change is right or wrong, nor what the fix should be. They simply make the change visible so that someone with more context can decide whether it is a regression, an improvement, or an acceptable trade-off.

The larger and more diverse that collection becomes, the easier it is to spot behavioural changes over time and understand where they come from. To me, that is part of maintenance rather than a maintenance burden.
`

vingerha:
We can have a lot of back/forth on this...trying to cut it short
My tests will focus on known issues/challenges and when I have a case I need to be able to create test quickly, not using any provider, hence this is my **risk-based limited** target for something that I am **willing to maintain**, I cannot but stress the maintenance part enough.
- synthetic only
- static-only route+local
- combined static-realtime, route+local but for departures only, no vehicle or alerts so also no output files
- no db
- no config flow
 
Other tests are for sure welcome, but outside of my test-structure, hoping that the output is readable in case of issues. Even my own structure was very poor at the start on the error-output, now just 'poor' but at least is is one that I know so I can analyse ...and will improve over time

Pulpyyyy:
A dataset, fixture collection, or specific test suite goes through the test tool and produces a result.

If the result is OK, then everything is fine. If the result is not OK, the test result becomes the input of the diagnostic tool, whose purpose is to determine whether the issue originates from the code, the data, or an assumption that is no longer valid.

Known issues, previous findings, and reference datasets can also be provided as inputs to the diagnostic tool. In practice, many of those datasets originate from users reporting unexpected behaviour.

Over time, these datasets can be reused by both the test tool and the diagnostic tool, gradually forming a valuable collection of reference cases and edge cases alongside the synthetic scenarios.

```                  +------------------+
                  | Dataset / Fixture|
                  | Test Suite       |
                  +--------+---------+
                           |
                           v
                  +------------------+
                  |    Test Tool     |
                  +--------+---------+
                           |
                     Test Result
                           |
            +--------------+--------------+
            |                             |
          OK                           Not OK
            |                             |
            v                             v
         Done                    +------------------+
                                 | Diagnostic Tool  |
                                 +--------+---------+
                                          |
                                          v
                                  Root Cause Analysis
                                   - Code issue
                                   - Data issue
                                   - Invalid assumption

Known issues ----------------------------+
Previous findings -----------------------+
User-provided datasets ------------------+

                                          |
                                          v
                           Reference & Edge Case Collection
                                          |
                                          +--> reused by Test Tool
                                          |
                                          +--> reused by Diagnostic Tool
```
Test failures generate valuable diagnostic inputs, while diagnostic findings generate valuable future test cases. When the feedback loop is shared, both capabilities improve together; when they are isolated, part of that learning effect is lost.

vingerha:
I know, I have instructed many testers but is a matter of cost/benefit and I just donot see a return to make it this beautiful, it is imo overkill plus (yes) maintenance. As you are aware, this is all unpaid work and not intended as a professional tool, I gain 0 out of it, just putting effort in...which is fine but also has its limits. 
What could change my mind is measuring the actual use, is it used by 10-50-200 or more people??? .... and user's feedback, but I have not yet found a way on how to achieve this....maybe one can trigger a HA notification upon an update of the code?

From where I currently stand:
When I have a proper set of tests, this will immediately help me after I apply a code change because if tests fail, then I already have sufficient lead on where to look. 
Although you recently provided a lot of good changes, I donot expect to see this continue at this rate ... and if the changes are applied in small(ish) chunks, tests will prove sufficient...still: my opinion :)

Pulpyyyy:
Replaced by #206, in tests_provider/ as asked.

PR #206, opened 2026-09-04
Add the provider tree: frozen feed fixtures and the journeys suite

Pulpyyyy:
Replaces #204 and #205, in tests_provider/ as asked on #205. Separate from tests/: it needs pygtfs, sqlalchemy and the protobuf bindings for real, and its cases come from real feeds cut down to a few routes. Only tests/ha_stub.py is shared, through a conftest that puts tests/ on sys.path for that one import; `pytest tests/` and its workflow never see the tree, test_provider.yaml runs it on its own requirements.

fixtures/: five feeds (GVB, Palm Bus, SNCF with trip updates and alerts, SNCF and TAO journeys), each with a manifest saying where the bytes came from, what was kept and why, plus the two scripts that build one from any zip or realtime url. test_journeys.py: for each route and direction, one test per promise (the stop list, a valid pair, the swapped pair) through get_stop_list and get_next_departure as they are, on a db built from the zip and reshaped the way pygtfs stores it, the clock pinned by freezegun to a day the trips run. The cases known to fail on main are marked xfail with the reason, strict, so a fix has to lift its mark. instructions.txt says how to run it.

Verified here on your main, in a clean venv with only tests_provider/requirements.txt installed; your tests/ stays green next to it.


vingerha:
Thanks, I copied it locally and this is what I see and how I compare it with my 'tactic'
First of all, cudos on the effort as what your code seems to tackle is maybe the most complex part of it all, db side and queries ... I threw it out just because of too many data issues with the various feeds I have seen and the complexity in creating proper input and reliable, verifiable output.

1. the on-screen output description is limited, and I cannot find detailed output (maybe I am missing things?), I have no clue where to look for what triggered below (example) 
```
[palmbus-A-d0-swapped] XFAIL (the swapped pair answers a departure the line does not ride)
[sncf-journeys-P8(A594575:)-d1-pairs] XFAIL (a train journey is matched by stop name on any line, not the asked one) 
```
Guesswork would be that this a data-only issue and not code ... but no clue how to analayse this. 

My alternative is pure data-driven, please look at below provoked error in 'results.txt', this does not at all interpret results so this is the other end of the test feeback and has its own form of vagueness. Here you need to be fully up to speed with the data in order to understand why it throws an error and forced to read a quite poor feedback format, e.g. if there are a lot of errors (which I had when creating the cases), the only option I had for a more visual overview, was to copy both sets in notepad++ and compare...not very user friendly.
I have not found a way to be in the middle of yours and mine (yet), Claude proposed a attribute-level analysis but this then means to copy code elements in the test. For me, since I constructed the tests/data and since I understand the data part, the results.txt helps me. No clue how to fix this challenge.

[results.txt](https://github.com/user-attachments/files/31858932/results.txt)

2. Your test file contains interpretation of the core code 
If the core code changes, the tests may still render without error although it no longer applies. 
Some queries are not complete, e.g. service_date only covers the calendar_dates
My approach is to lean fully on the code itself, hence already needed to rewrite gtfs_helper to get in-between datasets out. And with that approach there are limits too.

End result...both your and my attempts have pro's and con's and likely this is all due to the code the way it is constructed, it is hard to test.

Pulpyyyy:
Thanks for running it. On your two points, one is a presentation fault on my side, the other I think we are closer than it looks.

1. The detail exists, I just did not tell you where. `pytest tests_provider/ --runxfail` runs the marked cases as normal tests and prints, per pair, what was asked and what came back. Your two examples give, on your main:

```
the swapped pair GSNF0A -> CCO15A answered a departure the line does not ride
La Chaumusse - Fort-du-Plasne -> Champagnole: got La Chaumusse - Fort-du-Plasne -> Champagnole Paul-Émile Victor on P8
```

Both are code, not data. The bus query has neither route_id nor direction_id in its WHERE, so a sensor set on line A direction 0 is answered by a direction 1 trip of the same stops (trip 1418363764878 here). For trains the stops are matched with `stop_name like 'name%'` and no line filter, so Champagnole picks Champagnole Paul-Émile Victor, and a P8 sensor gets a K8+ train on the same stops. The fixtures are the published feeds cut down to the listed routes, the manifest says what was removed.

Why you had nothing to go on: I wrote instructions.txt from the writer's side. The marks were placed after I had read each failure, so their one-line reason is my diagnosis in short, and the file only describes the green run and the day a fix lifts a mark, not someone meeting an XFAIL and wanting to see what happened. Same for the messages: the line and direction sit in the test id (palmbus-A-d0-swapped) and are not repeated in the message, so read alone it says nothing. I will add a short "reading a failure" section to instructions.txt (what the id encodes, -rx for the reasons, --runxfail for the detail), make the messages name the asked line and direction and the trip that answered, and give the reasons in KNOWN the same precision. That is the only code I plan to add.

2. On reading the core: the test calls get_stop_list and get_next_departure unchanged, it does not re-implement them. The SQL it carries reads the feed, not the query: which directions a route has, the stop patterns of its trips, a day they run. That is the oracle. A wrong answer is one that contradicts the feed, so a change of meaning in the core cannot stay green unless the new answer also matches the feed, in which case it is right. service_date is incomplete as you say. The five fixtures are calendar_dates only so it never bit, and on a calendar.txt feed it fails loud ("no service date for a pattern"), not silently green. I will extend it with the first calendar.txt fixture.

I did not go through _fetch_departure_rows and _interpret_departure_rows by choice: the promise is on what the sensor shows, so it survives the next re-split of the helper, where results.txt is tied to the current one. I see the two as complementary, yours catches a changed row, this one a wrong answer.

Last thing, the reason it is built this way: the same tree runs on any checkout. The strict marks are the baseline of main. On a fix branch the cases it fixes turn XPASS and fail until their marks are lifted, so the list of lifted marks is the list of what the fix changed, per route, direction and pair. No compare script needed for that.

vingerha:
ok...
So the query results are interesting indeed. My request is to verify (with Claude, althouhg it freaked me out a lot) ... if you can use the pure code from the py scripts and not copy bits into the testscript. If it requires gtfs_helper.py to change and not enormously then let me know the approach but I do see that current code has limitations in certain test strategies.

On the yunthetic ones I actually do want to go through interpret rows as it had quite a few inconsistencies (and maybe still has a few), it is complex to read so any change should survive. For the static only cases, there is not much difference between interpret output and the coordinator output, I wanted the second one in preparation for adding the realtime, which the coordinator handles. Depending on the case, sometimes things need to be changed in SQL, interpret or coordinator, as I am skipping sql (for now), the other two are required.

I will be finishing 3 route/rt combined cases today and then my inital throw for 'route-based' is done. Next on the list will be localstop which has its own peculiarities (nightmares) esp. in combination with realtime.

vingerha:
Thanks a lot, I hope my maintenance fears are not becoming real (same goes for my own tests)

vingerha:
I ran the test for the details...maybe still missing things but if not the case, can you create a results_provider.txt or something with the errorneous elements? it is unreadable for me on the terminal with small print, colors and all

EDIT: Ideally, if feasible, the nicely worded first and then the details ... but I can live with the details alone as well

Pulpyyyy:
multi line txt is nice for reading but rather complicated to mass analysis/compare ... I think txt and json output vwould help

vingerha:
Yes but... donot overdo it on beautification. In a (almost) ideal world there is no error, then... if you change the code and an error pops up you have sufficient lead and if it takes a bit of time in case of rare errors, then it is fine for me it if is not too nicely formatted...as long as the details are there

Pulpyyyy:
Pushed 8d87d35 to my fork, but #206 was already merged when I did, so it comes as #207, rebased on your main. It is more than the messages I announced. Every check is now kept as a record, and a session hook writes two files next to the tree at the end of a run. results.txt has the shape of yours: one entry per case, and for a case that did not pass, the lines that broke the promise, what was asked and what came back. results.json holds every check of every case as fields, keys sorted: run the suite on two branches and a diff of that file lists what changed between them, pair by pair. That is the "calculated next to the expected, for another tool" your tests/instructions.txt asks for. Both files committed in #207 are the run on your main, as examples of the output. instructions.txt says how a case is built and how to read a failure.

PR #207, opened 2026-09-05
Provider tree: say what was asked and what answered, write results.txt and results.json

Pulpyyyy:
Follows #206 and answers the results file request. The commit message says what changes; nothing in the assertions does. Both results files committed here are the run on your main, as examples of the output.

On your question of 08:23: the tests already go through the code as it is, get_stop_list and get_next_departure, nothing of gtfs_helper.py is copied into the test. The SQL the test carries reads the feed (which directions a route has, the stop sequences of its trips, a day they run), not the code; that is the expected side. So no change to gtfs_helper.py is needed for this tree.

### DIFF
--- a/tests_provider/conftest.py
+++ b/tests_provider/conftest.py
@@ -1,9 +1,23 @@
-"""Bootstrap for the provider tree: the Home Assistant stand-ins of
-tests/ha_stub.py, registered before any test module is imported.
+"""Bootstrap for the provider tree, and the two results files.
 
-tests/ is put on sys.path for that one import, so the two trees share a
-single stub and a single answer to "what does the integration read off
-Home Assistant". Nothing else of tests/ is read from here.
+The Home Assistant stand-ins of tests/ha_stub.py are registered before
+any test module is imported. tests/ is put on sys.path for that one
+import, so the two trees share a single stub and a single answer to "what
+does the integration read off Home Assistant". Nothing else of tests/ is
+read from here.
+
+At the end of a session the records each case handed over through
+record_property are written next to this file, the way tests/ writes its
+own results.txt:
+
+    results.txt   one entry per case with its outcome, and for a case that
+                  did not pass the lines that broke the promise; to read
+    results.json  every check of every case, passed or not, as fields with
+                  the keys sorted; to diff between two checkouts or load
+                  into a tool. Written by a plain full run only, so a -k
+                  or --runxfail run does not overwrite a full one
+
+Both files in the repo are the run on main, as examples of the output.
 """
 from __future__ import annotations
 
@@ -15,3 +29,77 @@ sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "tests"))
 import ha_stub  # noqa: E402
 
 ha_stub.install()
+
+
+import json  # noqa: E402
+
+HERE = Path(__file__).resolve().parent
+_reports = []
+
+
+def pytest_runtest_logreport(report):
+    if "test_journeys.py::" not in report.nodeid:
+        return
+    if report.when == "call" or (report.when == "setup" and report.failed):
+        _reports.append(report)
+
+
+def _outcome(report):
+    if report.when == "setup":
+        return "error"
+    if hasattr(report, "wasxfail"):
+        return "xfailed" if report.skipped else "xpassed"
+    if report.failed and str(report.longrepr).startswith("[XPASS(strict)]"):
+        return "xpassed"
+    return report.outcome
+
+
+def _reason(report):
+    if hasattr(report, "wasxfail"):
+        return report.wasxfail
+    text = str(report.longrepr)
+    if text.startswith("[XPASS(strict)]"):
+        return text[len("[XPASS(strict)]"):].strip()
+    return None
+
+
+def _cases():
+    cases = []
+    for report in _reports:
+        props = dict(report.user_properties)
+        case = {"id": report.nodeid.split("[", 1)[-1].rstrip("]"),
+                "outcome": _outcome(report), "reason": _reason(report),
+                "checks": props.get("checks", [])}
+        case.update(props.get("case", {}))
+        if not case["checks"] and report.longrepr is not None:
+            case["error"] = [line[2:].strip() for line in
+                             str(report.longrepr).splitlines()
+                             if line.startswith("E ")] or [str(report.longrepr)]
+        cases.append(case)
+    return sorted(cases, key=lambda c: c["id"])
+
+
+def pytest_sessionfinish(session, exitstatus):  # noqa: ARG001
+    if not _reports:
+        return
+    cases = _cases()
+    lines = [f"test_journeys.py case results -- {len(cases)} case(s)", ""]
+    for case in cases:
+        lines += [f"case: {case['id']}", f"  result: {case['outcome'].upper()}"]
+        if case["reason"]:
+            lines.append(f"  reason: {case['reason']}")
+        broke = [c["text"] for c in case["checks"] if not c["ok"]]
+        broke += case.get("error", [])
+        if broke:
+            lines.append("  detail:")
+            lines += [f"    {text}" for text in broke]
+        lines.append("")
+    (HERE / "results.txt").write_text("\n".join(lines), encoding="utf-8")
+
+    option = session.config.option
+    partial = option.keyword or option.markexpr or option.runxfail
+    if partial or exitstatus == 2:  # a -k, -m or --runxfail run, or interrupted
+        return
+    (HERE / "results.json").write_text(
+        json.dumps({"cases": cases}, indent=2, sort_keys=True, ensure_ascii=False)
+        + "\n", encoding="utf-8")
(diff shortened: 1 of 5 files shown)


Pulpyyyy:
For information, the same tree run against my fork branch ext/rt-per-source (e4a34d8), the lots I have been sending you:

```
59 passed, 36 failed
```

The 36 failed are exactly the 36 cases marked known-to-fail on main: on the fork they pass, and a strict xfail turns an unexpected pass into a failure. No case that passes on main fails on the fork.

results.json against the committed one (the run on your main), check by check:

| | |
|---|---|
| unchanged | 738 |
| fixed on the fork (false to true) | 60 |
| broken on the fork (true to false) | 0 |
| right on both sides, different answer | 70 |
| GVB d1 cases with fewer checks | 9 |

The 70 different answers are another trip or another platform of the same station, five swapped pairs answered with nothing instead of a trip of the other direction, and TAO 22 d0, where main answers a direction 1 trip to a direction 0 sensor and the fork a direction 0 one; the pairs promise does not check the answered direction, I may tighten that later. The nine GVB d1 cases have fewer pairs because direction_repair moves trams 1, 7 and 17 to their real direction.

To run it on any branch and read the difference, from the component root:

```
git remote add pulpyyyy https://github.com/Pulpyyyy/gtfs2 && git fetch pulpyyyy
git checkout pulpyyyy/ext/rt-per-source
git checkout pulpyyyy/test/provider-results -- tests_provider tests/ha_stub.py
pip install -r tests_provider/requirements.txt
pytest tests_provider/ -rX
git diff --stat tests_provider/results.json
git diff tests_provider/results.json
```

The second checkout drops the tree and its committed results.json onto the branch; the run rewrites results.json and git diff shows what the branch changed, pair by pair. -rX lists the cases that passed against their mark. Nothing is committed: git reset --hard, then check out your branch again, and it is gone.

vingerha:
I noticed that too, so why do they pass on the fork and not locally ? or is this now fixed with this PR?

Pulpyyyy:
Passing on the fork because the fork carries the fixes: none of them is a PR yet, they only live as branches on my side. On the fork they show as failed only because their marks are still there.

What lifts them, one small PR each, marks removed in the same commit:

| Marks | Fix | Branch | PR | On your main |
|---|---|---|---|---|
| 14 swapped | route and direction in the WHERE of get_next_departure | fix/next-departure-direction | not yet | needs a rebase |
| 4 train pairs | match the station on the exact name the flow offered | fix/train-station-name-collision | not yet | applies as is |
| 18 stop_list | platforms of one place grouped (#198) | fix/stop-platform-groups | not yet | applies as is |
| 18 stop_list | same-name stops told apart by their station (#198) | fix/homonym-stop-labels | not yet | applies as is |
| 18 stop_list | ordering by the longest trip (#198) | inside my larger flow branch | not yet | has to be taken out on its own first |
| 10 GVB stop_list | direction_id repaired at import for trams 1, 7 and 17 | feat/direction-repair | not yet | needs a rebase |

You asked for fewer/smaller PRs ;)

vingerha:
ah, the test run action is not acting on xfail, I now see that... need to test mine too then

vingerha:
I did a quick check on palmbus since I have that working locally, the tests say that there is a duplicate stop_name but I have a different result, the script states route 22 but that one has no duplicate names with the same (!) lat/lon. In fact there are only a few for palmbus and they reoccur in most (all?) trips for their routes
```
SELECT 
    r.route_id,
    t.trip_id,
    s.stop_name,
	s.stop_lat,
	s.stop_lon,
    COUNT(*) AS occurrences
FROM routes r
left JOIN trips t ON t.route_id = r.route_id
left JOIN stop_times st ON t.trip_id = st.trip_id
left JOIN stops s ON s.stop_id = st.stop_id
GROUP BY r.route_id, t.trip_id, s.stop_name, s.stop_lat, s.stop_lon
HAVING COUNT(*) > 1
ORDER BY r.route_id, t.trip_id, s.stop_name;
``` 
A quick analysis learns that "Résistance Coubertin" has 2 different stop_id and the only difference is that one stop has weelchair_boarding and the other not...beets me how this looks in reality on site :) , but is it not a code issue.
I would not know how to deal with this when selecting stops

EDIT: unless you can revert my findings, from my pov this is what I disover often in user-issues. It is quite hard to test for things like this and then one needs to decide if the code needs to filter or not, in some case it should, in others maybe not. There are also temporary stops sometimes 50+ meters away from the temporary-blocked stop. And to get back ot A'dam GVB, they have a nightmare setup, so says also the NL overall gtfs data manager

vingerha:
And with "it is not the code" ... I mean that my current config_flow code is still not good for selecting things... this is the eledged dual stop, I donot see much difference for embarking a wheelchair LOL
image

Pulpyyyy:
I'm not totally aligned

Résistance Coubertin is not on the 22 and not in the fixture. In the full feed it is on BOC, Bocca Cabana: a loop of 18 stops. Two records because the loop starts and ends there (TAO Line 22 does too). The tree keeps both on purpose: two records of a station far apart in the ride are a loop, only consecutive ones are folded. Not a code issue, agreed.

The 22 d0 mark says a stop_id is repeated, not a stop_name, and your SQL cannot see it ? (i guees) ? it looks inside one trip, the doubling is across trips, in the union the selector builds. On your main get_stop_list("22", 0) returns 66 entries for 39 stops. Direction 0 has two patterns, the full run from Le Trayas (39 stops) and the short run from Théoule Plages (27 stops) renumbered from 1; the DISTINCT is on (stop_id, stop_sequence) ordered by sequence, so every stop of the short run comes out at both numbers, interleaved:

    THEC0A: Théoule Plages (1)      ...      THEC0A: Théoule Plages (13)
    ISL22R: Le Princes des Iles (2) ...      ISL22R: Le Princes des Iles (14)
    GSNF0R: Gare SNCF de Cannes (27) ...     GSNF0R: Gare SNCF de Cannes (39)

Direction 1 is clean because its short run leaves from the same origin as the long one. Same as the TAO tram B interleave from #198, on your own feed. That one is code: walk the fullest trip first, then slot in the stops it skips. It is the stop_list fix I mentioned as still needing to be taken out on its own.

My fault that you had to hunt: the message should name the repeated stops and their two numbers. I will make it in #207.

On the action: xfail is green by design, that is the mark doing its job.

Pulpyyyy:
One more thing on the 22 fix, walking the fullest trip first: it lives in commit [1e6f938](https://github.com/Pulpyyyy/gtfs2/commit/1e6f938f774f498e5072787784038774a40b7eec) of feat/two-database-flow, the config flow rework, and has to be taken out of it before it can come as a PR. get_stop_list is self-contained, about fifty lines, so the extraction is short. The platform fold alone, commit [de82e49](https://github.com/Pulpyyyy/gtfs2/commit/de82e49a6ebd6eba2b9eb05a3dc4c831cef6afff) on fix/stop-platform-groups, does nothing for the 22, I checked: it only merges consecutive records of one station, and these twins sit twelve stops apart.

Pulpyyyy:
Output fixed and pushed. A stop list failure now names what it judged, so it can be looked up in the feed without a query of your own:

```
the stop list offers a stop twice: THEC0A Théoule Plages at 1 and 13, ISL22R Le Princes des Iles at 2 and 14, THEG0A Théoule Gare at 3 and 15, and 24 more
the list offers one station twice in a row: ORLEANS:StopArea:TVERNE2 Jules Verne then ORLEANS:StopArea:TVERNE1 Jules Verne
a trip serves a stop the list does not offer: RPES1R Rond Point de l'Espace (on the ride CCO15A .. GSNF0A)
```

The same identifiers go into results.json as fields, so a diff between two branches reads stop by stop. Both files are regenerated, still the run on your main: 59 passed, 36 xfailed, no case changed side.

vingerha:
So, the earlier query runs on stop_name and I added lat/lon to make it really unique
If I use only stop_id (wihtout lat/lon) it returns nothing, i.e no duplicates **within a trip** 
```
SELECT 
    r.route_id,
    t.trip_id,
    s.stop_id,
    COUNT(*) AS count
FROM routes r
left JOIN trips t ON t.route_id = r.route_id
left JOIN stop_times st ON t.trip_id = st.trip_id
left JOIN stops s ON s.stop_id = st.stop_id
GROUP BY r.route_id, t.trip_id, s.stop_id
HAVING COUNT(*) > 1
ORDER BY r.route_id, t.trip_id, s.stop_id;
```
if you add stop_name to this query, then yes, it will find twice the name as the stop_id is unique but not the the combination of stop_id and stop_name

EDIT; and just to be clear, that it works for Palmbus is guarantee that it works in other feeds

vingerha:
with all this good progress I raised a poll, you likely have seen it... not sure if htis opens a box of pandora

Pulpyyyy:
Your query is right, and it cannot show this one: `trip_id` is in the GROUP BY, so it looks inside a trip, while the repeat is across trips, in the union the selector builds. Drop the trip and it appears:

```
SELECT stop_id, GROUP_CONCAT(stop_sequence) AS numbers
FROM (SELECT DISTINCT st.stop_id, st.stop_sequence
      FROM trips t JOIN stop_times st ON st.trip_id = t.trip_id
      WHERE t.route_id = '22' AND t.direction_id = 0)
GROUP BY stop_id HAVING COUNT(*) > 1;
````

Agreed that one feed proves nothing, which is why the same shape is in the TAO tram B case from #198, on your own feed.

Issue #181, opened 2026-09-01
[FEATURE]: Discussion Improve start/stop selection


vingerha:
### Describe the solution you'd like

Depending on the feed/provider, it is not always easy to build a route based on start/stop. In quite a few case it just cannot find a matching trip based on the selected combination so you need to try-and-try-and-try other combinations, and/or change the direction (as there is none in the feed or all 0), etc. I also have situations where there are multiple stops with the same name on the same route, beats me why but often only one is working in combi with the other. I thought about grouping them by trip but that is impossible for certain cases, the other solution is to have the end-stop be filtered on the start-stop chosen.

vingerha:
feedback @Pulpyyyy 
On the start/stop matching: I hit the same wall and reworked the config flow, on my fork only for now. The direction step no longer shows a raw 0/1 but a label built from the first and last stop of each direction. The stops are then listed in travel order for that direction, same-name stops (platforms, or a circular terminus) numbered in calling order, and a destination placed before its origin is rejected right at the form, so a pair that cannot match any trip never reaches the sensor. One cause is in mainstream as well though: the stop list query merges all trips of a direction ordered by stop_sequence, but partial trips renumber from 0. On my TAO feed, tram B has 110 short runs and the merged list interleaves, real positions come out as 0, 19, 1, 20, 2, 21... That misorder alone makes valid pairs look wrong and wrong pairs look valid, hence the try-and-try. My planned fix, not written yet, is to order by the longest trip of the route and direction; with the order right, your idea of filtering the end stop on the chosen start becomes reliable too. I can show you the flow rework after your release if you want.

vingerha:
It is a little more complex than that. 
Example below shows stops for tramline 1 in Amsterdam, direction outbound.
If I indeed want to depart from a specifc stop 1 then this is 'easy' but if I donot want to start from stop 1 and stop before the end, then which one to choose. e.g. stop 'Rijksmuseum' shows up 5 times in the list at different order#. 
Even when basing the end-stop on the first one, I would still not know which one of the 5 stops to select if I start at 'Rijksmuseum'. The only benefit of that approach is that you would be able to change start and see what 'end' stops are possible. 
For trains there is a similar issue where a large trainstation has multiple platforms and no clue which to select, hence this is now on 'name' only....so maybe that might be the solution, show only the unique stop-names on a route and then filter in the query 

[...]

PR #211, opened 2026-09-10
Offer each stop once, in riding order, then the destinations a trip really reaches

Pulpyyyy:
This is the stop selector improvement from #181 and #198, on its own: three commits, nothing about today/tomorrow, next service dates or the database.

**What changes**

- The stop list of a route and direction is walked from its fullest trip, so each stop appears once, in riding order. Sorting the union of all trips by stop_sequence interleaved the start of a route with its middle (TAO tram B, 110 short runs) and offered a stop once per number it was seen under (Palm Bus 22: 52 entries for 26 stops).
- Repeated names are numbered in calling order (#1, #2): a loop calls at its terminus twice under its own ids, a square carries four stops of one name. Same-name stops stay distinct entries; folding by name would show departures a rider cannot take.
- The flow asks for the origin first, then offers the destination among the stops a trip really reaches from it, so a pair no trip rides never gets to the sensor. The old retry screen goes; when nothing runs today or tomorrow the destination screen comes back with that said and the picks kept.
- Texts in strings.json and the five translations.

**Tests**

- tests_provider: a `destinations` promise joins the tree (from three stops of each sequence, every stop the trips through it ride to after it, once, in riding order, nothing else). Eleven stop_list marks are lifted; the seven left all fail on two platforms of one place offered one after the other, which is a separate change. 99 passed, 25 xfailed.
- tests/: 10 passed.
- The two screens were driven on the TAO fixture db outside HA: origin list, destination list, the screen shown again on `stop_incorrect` with defaults kept, the entry created, the train renaming, the `no_destination` and `no_stops` aborts.

Config entries are unchanged: the sensor still receives `stop_id: Name (sequence)` entries and queries on the id.

### DIFF
--- a/custom_components/gtfs2/config_flow.py
+++ b/custom_components/gtfs2/config_flow.py
@@ -56,6 +56,7 @@ from .gtfs_helper import (
     get_next_departure,
     get_route_list,
     get_stop_list,
+    get_destination_stop_list,
     get_datasources,
     remove_datasource,
     check_datasource_index,
@@ -75,6 +76,16 @@ TRANSLATION_DESCRIPTION_PLACEHOLDERS = {
     "docu_configuring_options": "https://github.com/vingerha/gtfs2/wiki/04:-Configuring-a-route's-options-(inc.-adding-real%E2%80%90time)",
     "model": "Example model",
 }
+
+
+def _stop_name_of(entry):
+    """The stop name inside a picker entry "stop_id: Name #2 (12)": the
+    number given to a repeated name and the sequence are display only."""
+    name = entry.split(": ", 1)[1].rsplit(" (", 1)[0]
+    if " #" in name and name.rsplit(" #", 1)[1].isdigit():
+        name = name.rsplit(" #", 1)[0]
+    return name
+
 
 @config_entries.HANDLERS.register(DOMAIN)
 class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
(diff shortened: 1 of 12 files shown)


vingerha:
Can you please run this PR against my branch first as I just finished that one.
In short on the branch (well...short?)
- I wanted the query to select anything in the future, not being limited by today/tomorrow, just the future of 'now', In the old/current situation, the query collected past departures for today too and this is no longer possible. The query return max 30 rows
- In order to do that I created a sql-list with valid dates, i.e. dates for trips that really exist. This is based on either their start/end stop (route-based) or their stop+stop_times (local-stop)
- error prone and poor-readable code that analysed whether a queried row belongs to today or tomorrow ... is fully gone
- no additional tables are generated
- due to these simplifications I slightly modified your tests_provider/fixture_db (e.g. now_date no longer exists)
- gtfs_helper code has shrunk more than 160 rows

Your provider- and my sythetic-tests all pass but (!) I cannot be 100% sure if all is covered...there may be some regression in a remote area. 

Two options: 
- either I push this branch first into a release and see how much issues come back
- or your selection fix is added first

Tbh, I prefer the latter as my fixes are invisible for the user and pushing a release without visible improvements or bug fixes is not ideal...

Pulpyyyy:
Ran it on your branch. Merged this PR into remove_today_tomorrow_mechanism, the only conflict is tests_provider/results.txt which you deleted, I dropped it.

Your synthetic tests: 14/14. Provider tests on your branch alone: 59 passed, 36 xfailed. With this PR on top: 99 passed, 25 xfailed. The 11 stop_list xfails lift (gvb 1, 7, 13, 17, palmbus 22 and B, tao 40) and the 29 destination cases pass, nothing else moves. So it holds on your branch, and I agree with your order, this one first.

Separately, one thing I hit on your branch that has nothing to do with this PR: departures after midnight are dropped. The query keeps a 24:03 trip thanks to the +1 day case in the WHERE, but it returns vd.date next to time(departure_time), so python rebuilds "<service date> 00:03", which is in the past, and skips the row. On the gvb fixture, trip 380786522, clock frozen at 23:58 Amsterdam: main answers 2026-09-27 00:03:30, your branch answers {}. No test case of either suite carries a stop time past 24:00, which is why nothing caught it.

If you want it, what I would suggest is to stop reading a GTFS clock as a time of day and lay it on the service day instead: two small helpers, seconds since the service day's midnight and midnight plus those seconds, the query returning the stored stop times rather than time(...) so the marker survives the trip to python, and then the rollover guessing in _interpret_departure_rows can go, both the "if arrival < departure then +1 day" chain and the combine(upcoming.date(), ...) block. It also makes the stop time attributes agree with next_departures when the agency zone is not the HA one, today one is stamped in each. I have it working here: your tests 14/14, provider 59 passed 36 xfailed unchanged, night trip answered, day still the service day. Say the word and I open it as its own PR on your branch, with a night case for the suites.

Pulpyyyy:
Opened it as #212, against your branch rather than main, so you can read the diff whenever you get to it. Nothing urgent, and it does not touch this PR.

vingerha:
That tests need to be adapted is no surprise.
What I read is that some provider tests failed AFTER the merge, correct?

Please let me know where you (or Claude) see that post midnight departures are dropped, my synthetic tests all pass with the code and have today only, tomorrow only, mixed and with realtime, so the above is a surprising finding but yes, it all depends on teh quality of the test data.  So if it is 'just' gvb then I will have a look at it because a lot of the code has (yes again) some case behind it that may or may not apply... above says at least that is seems wrong right now :)

Pulpyyyy:
Crossed messages ;)

vingerha:
the most questioning bits are that from the 230+ people approached in the 'poll' only 25 responsed/acted plus some not...so net result, < 40 seemed online/interested.... so is all this effort worth the output. For me it is spare time, so half-half ok and with ok I mean: it has to be near-perfect

Pulpyyyy:
I totally get the spare time aspect: we are definitely in the same boat there, which is why we want to keep things efficient (+backport for PR)

As for the low engagement, honestly, I wouldn't sugarcoat it: if only 40 people out of 230 care, it probably just means the current state of the core is too frustrating or hard to follow, and people have simply moved on. Making things rock-solid and intuitive is the only way to win them back.

vingerha:
Agree fully and a new option to 'market' the integration... repeating the thanks for any tests/pushbacks/etc/etc/ ...esp. the not always being positive stuff is (for me) a help ... :)

vingerha:
After your finding on tomorrow-trips, the code has now been updated
- route and stop-based configs now use the full stamp from the db and convert that to the actual departure date, no more fiddling if/then/else. The SQL code changed but hardly on performance, mostly below 50ms...but this is a QED with other feeds.
- all references to 'include tomorrow' have been removed as this is no longer an active attribute (config_flow/gtfs_helper/const/coordinator/sensor/strings/translations) 
- synthetic tests have been updated and run without error, I added but modified your use-case test-data to align labels
- provider tests did not need an update and also run without error

At this point, the only thing missing is the proper selection of stops
Question is: do you want to test the current branch or shall I add the stop selection first

EDIT: manual tests
- adding new stop based and new route-based entries, both line related as well as trains
- restarted existing entities, no errors

vingerha:
I spent a few hours now on verifying your code and it seems fine but the underlying data sucks at times. 
- gvb line one shows first start Azartplein and much later the other one so two sets of data in 1 dropdown
- zou 653 shows it is using the same 2 stops opposite of one another but for both directions so 2x2 combinations, aaargh

I am fed up with data disappointments today...will retry tomorrow :)

vingerha:
but...can you please resync the pr with the current branch, I had to copy config_flow and gtfs_helper changes it by hand due to a massive out-of-sync

EDIT: I have tried lots of options now to get to a 'nice' list of stops but it seems quite complex, also due to the data issues. At the moment I see an improvement by making destination-stops depending on start stops but I am wondering if there is a better way to show start stops, I feel like being in a rabbit hole, like AI :)

vingerha:
I am also wondering how other providers handle this, e.g. google maps or zou website. They donot seem to have these issues....or at least: not visibly

Pulpyyyy:
Rebased on remove_today_tomorrow_mechanism and switched the base to it, so the diff is only the 3 commits. The include tomorrow toggle is gone from the destination screen, as on your branch. Your tests 17/17, provider 99 passed / 25 xfailed, same as before.

Pulpyyyy:
On GVB 1, that one is the feed: about 40% of tram 1 trips carry the direction_id of the other way, so each direction lists both chains. No stop list can fix that at display time. My fork repairs direction_id at import, but that is a separate and bigger topic. Working on unrefined data means that sooner or later you end up writing corrections that should never have been needed, in every place that reads it.

On Zou 653, each of those places is two records under one parent_station, one per side of the road as you saw (40 of the 59 names on that line), and the two variants of the line use different sides. What would fix it is to treat a station as one stop: list it once, and let the departure query match any record of it. Then picking either one gives the same departures and the 2x2 goes away. My fork has the matching half; the listing half does not handle lines like 653.

On Google Maps or the Zou website: they never ask you to pick from a list of a line's platforms. You pick a place, it resolves to a station, and the line and direction follow from where you want to go. That is the side I already took: the origin is asked first and the destinations are filtered from it, on the trips that really leave from there, so it is not a matter of names alone. In my fork it goes down to the mode: on a line mixing trains and coaches, a train station only leads to the stations a train reaches from it, and a coach stop to the ones a coach reaches. Stations instead of platforms would go the same way.

Same goes for the config flow: fixing one screen at a time only goes so far, it is the flow as a whole that would need rethinking.

vingerha:
> On Google Maps or the Zou website:

Google maps, true, yes I 'forgot' it does not start with stops
Zou however is also allowing to select a stop too but I guess they then indeed treat the parent, then this likely tries to overcome sloppiness (? or why else would a two stops at 5m opposite side of the road serve two directions?)
```
stop_id                     direction_id    trip_count
ZOP:03108-0615200    0                  32
ZOP:03108-0615200    1                  11
ZOP:03109-0615200    0                  12
ZOP:03109-0615200    1                   9
```
For using the 'parent', I can no longer be sure but in the past only a few providers (that I have seen) contained a parent stop (area) and e.g. Rome or Renfe still do not but that may be because they donot need it .... it may be worth looking at.

Pulpyyyy:
On parent_station, it can only be a bonus, never the rule. Zou, SNCF and Palm Bus give one to every platform, the Dutch feed to most, but TAO has almost none. So: group when there is a parent, keep the stop alone when there is not.

I reproduced your 653, and it is not sloppiness. The line runs several itineraries each way, and two variants of the same direction do not call at the same pole. So direction_id is right and there is nothing to repair, the pair simply is one station. List the station, let the departure query match any of its poles, and the 2x2 goes away on its own, whatever the variants do.

The departure query in my fork already matches siblings on parent_station, so the missing half is only the listing. Existing entries stay valid, a child id still matches itself.

My preference is to keep #211 to riding order plus destinations filtered from the origin, and do the station listing right after, on top of it. Tell me if you would rather have it inside this PR.

vingerha:
You need to explain the zou part a bit better
ZOP:03108-0615200 is on one side of the road and serves 653 in 0 and 1 direction
ZOP:03109-0615200 is at the opposite side of the road (different lat/lon) also 653 and also 0 and 1
So 
- trip ZOP:1005289-1029114 in direction 1 serves ZOP:03109-0615200
- trip ZOP:1005290-1029115 in direction 1 serves ZOP:03108-0615200

i.e. one of the trips is supposedly stopping at the wrong side of the road

vingerha:
Did some digging and direction_id is supposedly 'cosmetic' value, no guarantee... this is a surprising new finding for me after all these years

Pulpyyyy:
I need to look at this properly, but not right now (need more coffee), I have another side to finish first. Nothing blocking here: #211 stands on its own, riding order plus destinations filtered from the origin, and the station listing comes on top of it later.

vingerha:
I will have a further look too but also busy on the other side (regular life).... 

But this finding does support quite a few of the issues in stop selecting, e.g. GVB shows the Azartplein-group and Matterhorn-group in the resultset for 'line 1 outbound' where Matterhorn is at the very end of line 1 looking from Azartplein.

Pulpyyyy:
Your GVB case is a different mechanism though, no? There the label is wrong and the poles are fine, in Zou the label is fine and the poles are duplicated.

So the repair differs. For GVB, classify each trip against the canonical stop order of the line and rewrite direction_id (at import, better imho). For Zou, group the poles under the station. Both come down to the same rule: trust the order of the stops, never the labels around them.

vingerha:
For zou, the listed stops and order is fine for outbound (dir_id=0), just confusing about the stops since they are only duplicates due to the different trips they serve, inreality this is 1 stop
For gvb 1, for outbound the list shows a group that start with Azartplein, ending in Surinameplein, which is correct. Then it adds a secondary group starting in Matterhorn and ending in Azartplein, so the exact opposite of the first one, which is confusing if one selects 'outbound' and likely this is supported by dir_id not being a guarantee, see map with poor labels :)
I have no clue if this behavior shows up with other providers or other routes

image

Trying to formulate what I want vs. what to expect.
If I use gtfs2 right now and actvely select outbound then I would not expect the Matterhorn group as it is in reverse order (or v.v.)
As a 'user' (thinking 'me' only, not sure about others), I am primarily interested in stop>stop and I would not really know when a route is outbound or inbound unless I know the route by heart.
if not being given the out/inbound option, then I would expect a list of stops in order end-2-end, in above case from Azartplein to Matterhorn (or v.v.) and I can select any stop, followed by  the (already existing) limited destinaton stops on the second screen.
Additional on selecting out/inbound: Palmbus B shows Mouans-Sartoux > Cannes for the route description but outbound is actually the other way round...confusing
Lots of TLDR ... maybe drop out/inbound all in all?

Additional idea, (much) later PR... have the app show the nearest stop first based on lat/lon of the device (beautification)

Pulpyyyy:
A loop would still need the direction, so we have to keep remembering it at least.

About map, we can nearly do whatever we want
image
Zoomed 
image
image

But inside config flow even thought it would be really nice, it's more a serious challenge (HA configflow is very restrictive in style/content)

Pulpyyyy:
If you have test env, give try to fork that would give an idea about UX/UI (ext/rt-per-source is the most adavance) with configflow

vingerha:
Sorry to ask but do you have an example for a loop somewhere, just feed/provider and route ?
Guessing about a loop, if 1 > 2 > 3 > 4 > 5 > 1  and if this loop is also served by 1 > 5 > 4 > 3 > 2 > 1 then indeed, if I am at '2' and need to go to '1' then I want the second (return) trip ... I had this challenge earlier in Halifax and google maps could not present a good picture then :) so maybe you have something already in your daily life?

Pulpyyyy:
TAO bus 22 is a loop

image
image

https://tao-mobilites.fr/fr/9o-Telecharger-plans-et-fiches-horaires.html#fiches
https://transport.data.gouv.fr/datasets/gtfs-netex-et-gtfs-rt-reseau-tao-orleans-metropole

vingerha:
Thx, maybe later, now back in prehistoric times with helping the son with suites , récurrence and hérédité (seems a lot like trips)

vingerha:
Spent another hour on this but not making much progress to get to a proper list just-like that, so please have a look at this
Conditions/given:
- The list with start stops must contain all stops on a route
- It is not possible to use direction_id to get to a stop list guaranteed in one direction (gvb and zou 653 example)
- stop_id is mandatory for further filtering as the name is not unique enough (zou 926 example)

Approach:
- use the overall context of your code but **drop the direction_id** 
- identify the trip with most stops and take that as basis
- add missing stops, where stops with a duplicate name have to be added if their stop_id is different, as there is no option to find out when and on which trip they are used (zou 653 example)
- make sure that added stops are in the right location/order, not sure how performance intense this is

What could be done for the final presentation is to try and still show unique names and use a supporting table when they belong to the same parent (parent not being null)

Back tomorrow

vingerha:
closed as superseded by https://github.com/vingerha/gtfs2/pull/215

PR #212, opened 2026-09-10
Lay every stop time on its service day, so a ride after midnight is kept

Pulpyyyy:
This targets remove_today_tomorrow_mechanism, not main.

A ride that leaves after midnight is dropped. The query keeps its row, the +1 day case in the WHERE sees to that, but it hands the row back as vd.date next to time(departure_time), so _interpret_departure_rows rebuilds `"<service date> 00:03:30"`, finds it in the past and skips it. On the gvb fixture, trip 380786522 with the clock frozen at 23:58 Amsterdam: main answers 2026-09-27 00:03:30, the branch answers {}.

A GTFS clock is not a time of day, it counts from the midnight its service day started at, so every stop time is laid on that day here instead of being read as a wall clock. _gtfs_seconds folds a stored value into the seconds the feed counted from that midnight, _lay_on_service_day adds them to it, and the query returns the stored stop times rather than time(...) so the marker survives the way to python. The rollover guessing then has nothing left to decide and goes with it: the chain comparing an arrival against a departure to know whether the day turned, and the block recombining an upcoming arrival on the departure's date.

It settles a smaller mismatch on the way. The stop time attributes were stamped in the Home Assistant zone while the departure list used the agency's, so a feed abroad reported one trip at two instants. With HA on America/New_York and a gvb feed, the same ride read 21:54Z in next_departures and 03:54Z in origin_stop_time. Both say 21:54Z now.

_gtfs_seconds also reads a plain HH:MM:SS, so your existing case files, captured when the query folded the time itself, are read unchanged. Nothing had to be recaptured.

tests/case_route/case_5 is new: a ride crossing midnight on the way, one leaving after it, and the first ride of the next service day, all asked at 23:58. It fails without this change and passes with it. Before it, no case of either suite carried a stop time past 24:00, which is why nothing caught this.

On your branch with this on top: tests 15/15, tests_provider 59 passed and 36 xfailed, unchanged.


vingerha:
Thanks for the discovery and I now understand the issue. However, I have just started to remove python code to make it easier for each step to read and this PR re-introduces python complexity with additional functions. 
To be a bit more clear on my view:
1. sql code kicks out proper rows, each with a undeniable time,
2. interpret results, should resolve bits and pieces 
3. coordinator picks up from the interpreted results and adds realtime

The whole idea is that I had dozens of issues to look at sql and then analyse how they transform.

That was the idea but my branch was clearly still not fully in line as 1. was incomplete/correct, and 2. still has complexities
So, I will change the sql as this is at the source and then use bits of this pr in the 'interpretation'
The sql runs ok on performance and I can use that output later in other tests, i.e. I can edit a sql output much easier if I know it represents real trips and stops etc.

vingerha:
in between update, even with the sql part, the interpret remains harldy unchanged, sigh... tomorrow is a new day...still THANKS

Pulpyyyy:
Is forcing everything into SQL really a good strategy? Isn't Python inherently much more flexible for handling data manipulation and complex logic? SQL excels at what it was built for fetching, filtering, and joining large datasets at the source but trying to push temporal calculations, edge cases, and business logic entirely into queries usually just turns into an unmaintainable puzzle. Python shines precisely where SQL hits a wall: readability, step-by-step debugging, and adapting to messy real-world data. When a project tries to strip out Python in favor of pure SQL, aren't we just trading one kind of complexity for another and making future fixes much harder to reason about?

As now the DB can be stripped down to only the needed data, I'm not even sure SQL is better suited for this anymore.

vingerha:
it is either the complexity is in the SQL or in the python code and although the gtfs_seconds etc. seem good and likely stable, in case of issues it requires debug code per row to see what happens where.

My preference is to have the output of each step properly readable and applicable. This case is exactly a reason why as from the sql output it previously looked all fine but the interpretation function falsely removed things, which I could not read easily from the various outputs. You already mention step-byt-step debugging but this can become very annoying
**As long as the query does not suffer from important performance decay**

So in short my 'logic'
- sql exports the departures in a (now=new) date-time that applies locally, with the sql output in a db viewer you can already verify the rows with the provider timetables
- interpret_rows converts it to a timezone proper setup, this since the TZ can come from multiple locations. Added, in case of stop-based config it runs through each row to extract a RT value... still not changing the static departires.... in the end it converts it to a dict that is usable in coordinator.
- coordinator adds addtional actions, e.g. for route-based it adds the realtime data and feeds it to the sensor

vingerha:
Closing it as it is fixed in a different way

PR #213, opened 2026-09-11
Remove today tomorrow mechanism with stop selection

vingerha:
@Pulpyyyy 

I am a bit (?) stubborn as to keeping things simple so I returned on my idea to have the start / end selecton based on name only
This branch has all in it, it is based on the other branch (Remove today tomorrow mechanism)

Why  would I want this?
- it reduces the stop to just stops, not stop_id... when using stop_id or even lat/lon, as a user I have no clue which one to take as a start (e.g. line 1 Rijksmuseum). The current setup just displays all possible stops and during the destination selection it will find all other options. The query then finds all possible trips between the stop_names.

Most of the provider tests fail but that is expected.... what I would really like from you is to come with critical arguments.
The code allows to use existing start/stop and worked with me for all 5 route-based test entities


vingerha:
Adding: whilst running through your test cases I discovered things, some are easy to fix in the test case but others seem to be with the simplified new code ....working on it.

Pulpyyyy:
Simple is good, but not if the sensor then shows departures from the wrong stop. A stop name is a label, not a place. Cases where name-only gives a wrong answer:

- Zou (Région Sud), line 926 Banon - Apt: up to five stops are called "Centre" and two "Ecole", each in a different village (Revest-du-Bion, Viens, Caseneuve, Saint-Martin-de-Castillon, Saignon). Zou prints the village only on its paper timetables, not in the GTFS. With name-only, "Centre -> Apt" gives the bus leaving whichever Centre comes first, kilometres from where the user stands.
- TAO Orléans, line 22: two different stops share a name 200 m apart (00049161 and 06001010). I tried matching on names in the fork: the query answered from the other one, and our offline suite went from 60 to 149 failures.
- Loops: line 22 runs Zénith to Zénith and calls at "Zénith" three times. The name cannot say which call you board.
- GVB tram 1: the two "Surinameplein" stops belong to two different stations, one of them filed under Hoofdweg.
- Stops get renamed. TAO renamed stops in 2023: an entry that stores the name stops matching the day the name changes.
- You said it yourself on #181: grouping all on one name may be wrong if the stops are 100 m apart.

What name-only is meant to fix (Rijksmuseum listed five times, a stop split over two platforms a few metres apart) is fixed without names: #211 lists each place once, and matching the stop together with the stops sharing its parent_station catches every platform. Where names are right, that gives the same answer; where they are not, it does not break.

For trains, names are fine, a station name is unique, but only as an exact match: the prefix match still on the branch (LIKE 'name%') boards you at any station whose name extends the one asked for (Champagnole and Champagnole Paul-Emile Victor).

vingerha:
thanks for this, ... then maybe this is not the way forward :) will review **now**

vingerha:
So, analysing the zou one is sufficient to show a unresolvable gap .... close (for some feeds) but certainly no cigar :) 

THANKS :)

vingerha:
btw, trains will need a review too, when I select Nice as end-station, a LOT of trips/options pop-up in the entity, one needs to enter soemthing more specific like Nice-Ville . The current implementation provides no help to identify that, this is for a later PR

PR #214, opened 2026-09-11
Keep the previous evening's late trips after midnight

Pulpyyyy:
After midnight, a trip timed past 24:00 by yesterday's service is still to come, but both queries start their days at today. The local stop's candidate_dates only holds today and tomorrow, so at 00:10 the 00:22 tram of yesterday's service is not listed. On the route side, cal_expand starts at max(start_date, today): calendar_dates feeds keep those trips because that branch has no lower bound, a feed using calendar.txt loses them. The fix adds yesterday to both, the time window still drops what already left.

The first commit adds tests_provider/test_night.py. For every fixture with calls past 24:00 (gvb and sncf) it asks at 23:50, and just before the call the morning after, and compares with what the feed has. Each case also runs on a copy of the db with the services written as calendar rows, since no fixture uses calendar.txt. Your route query reads SQLite's own clock and zone, so the test pins TZ to the agency's on Linux.

On your branch 5 of the 6 cases fail, with the fix they all pass. Your tests/ stay 17/17, the provider journeys are unchanged, and it merges cleanly with #211. I left results.txt and results.json out to keep the diff small, a full run adds the 6 cases.


Pulpyyyy:
Added a calendar.txt fixture, since none of the existing ones use that table: Adelaide Metro, the Glenelg tram (weekday, Saturday and Sunday services with removals and additions) and night bus N1, whose Saturday-only service runs into Sunday. The night cases pass on it with the fix.

Reading it needed two fixes in the test tree. fixture_db stored calendar's weekday flags as text where pygtfs stores integers, and a text '0' is true in Python and never equals 1 through a CASE. service_date only read calendar_dates. build_fixture.py also gets --keep-night, the per-pattern cap dropped the night runs.

Its three stop_list cases are marked like the other selector ones, they pass with the selector of #211.

vingerha:
Thx, is this claude review or did you put your brains into it ?

Pulpyyyy:
I struggled with the times/midnight transitions on the TAO A/22 and N lines. But Claude gave me a hand with running the tests in bulk and finding other, more convincing examples.

I always try to be the brains of the operation, double-checking what Claude analyzes and/or suggests. He’s just faster than me when it comes to writing things up (he’s unbeatable at documentation, but I still have the edge when it comes to algorithms) and synthesizing the data.

PR #215, opened 2026-09-13
One stop list per line, and the pair of stops decides the direction

Pulpyyyy:
Replaces #211: its three commits are here, rebased on #214. Merge this one only, #211 can be closed.

**Stop list.** One list for the whole line, each place once. Records under one parent_station are one place, or, without parents, records with the same name within about 150 m (TAO). The order comes from the trips' stop_sequence. direction_id only decides which end the list starts from: the way most direction 0 trips ride it.

**Where the bus goes.** After the departure stop, the flow asks only when buses from that stop go to different termini, and offers each terminus. On a loop the next stop is shown with it (TAO 22 from Zénith: Plissay or Jean Moulin). At the end of a line nothing is asked. It is built from the trips, so wrong direction labels (GVB 1) hide nothing.

**Destinations.** Only the places a bus really reaches that way, one branch at a time, busiest first (Zou 653 from Grasse).

**Departures.** Both ends match the whole place, on the entry's line only, and a trip passing a place twice counts its shortest ride (Palm Bus 21 at Gare SNCF). Each departure says which stop it leaves from in next_departures_origin_stop_id, and the station attributes follow it. On a loop terminus the rotation is stored as loop_direction, the only direction the query reads.

**Upgrade.** No migration. Older entries keep working, but now only show their own line.

**Tests.** Two checks of tests_provider read differently now: stop_list is one list for the whole line, each place once, and a swapped pair passes when a trip really rides it. So the SELECTOR and SWAPPED marks are lifted, the TRAIN marks stay. No case removed, destinations and towards cases added: 179 passed, 4 xfailed. tests/ records next_departures_origin_stop_id and loop_direction: 17 passed.

### DIFF
--- a/custom_components/gtfs2/config_flow.py
+++ b/custom_components/gtfs2/config_flow.py
@@ -37,6 +37,7 @@ from .const import (
     CONF_ROUTE_TYPE,
     CONF_ROUTE,
     CONF_DIRECTION,
+    CONF_LOOP_DIRECTION,
     CONF_ORIGIN,
     CONF_DESTINATION,
     CONF_NAME,
@@ -57,6 +58,7 @@ from .gtfs_helper import (
     get_route_list,
     get_stop_list,
     get_destination_stop_list,
+    get_pair_direction,
     get_datasources,
     remove_datasource,
     check_datasource_index,
@@ -324,7 +326,6 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
                 data_schema=vol.Schema(
                     {
                         vol.Required(CONF_ROUTE, default = ""): selector.SelectSelector(selector.SelectSelectorConfig(options=route_list, translation_key="route",custom_value=True)),
-                        vol.Required(CONF_DIRECTION): selector.SelectSelector(selector.SelectSelectorConfig(options=["0", "1"], translation_key="direction")),
                     },
                 ),
                 description_placeholders=TRANSLATION_DESCRIPTION_PLACEHOLDERS,
@@ -333,12 +334,15 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
         user_input[CONF_ROUTE_TYPE] = user_input.get(CONF_ROUTE).split('##')[0] 
         user_input[CONF_ROUTE] = user_input.get(CONF_ROUTE).split('##')[1].split(": (")[0]   
                
+        # no direction is asked: the rider picks where they are and where they
+        # go, and the order of the stops on a trip says which way that is
+        user_input[CONF_DIRECTION] = None
         self._user_inputs.update(user_input)
         _LOGGER.debug(f"UserInputs Route: {self._user_inputs}")
         return await self.async_step_stops()
 
     async def async_step_stops(self, user_input: dict | None = None) -> FlowResult:
-        """Pick the origin: every stop the route rides in this direction."""
+        """Pick the origin: every place the line rides, both ways round."""
         errors: dict[str, str] = {}
         if user_input is None:
             try:
@@ -346,7 +350,7 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
                     get_stop_list,
                     self._pygtfs,
                     self._user_inputs[CONF_ROUTE],
-                    self._user_inputs[CONF_DIRECTION],
+                    None,
                 )
                 if not stops:
                     raise ValueError("no stops")
@@ -387,6 +391,16 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
             if data[CONF_ROUTE_TYPE] == '2':
                 data[CONF_ORIGIN] = _stop_name_of(data[CONF_ORIGIN])
                 data[CONF_DESTINATION] = _stop_name_of(data[CONF_DESTINATION])
+            else:
+                # the pair says which way the rider goes, except with a loop's
+                # terminus at one end: then the entry keeps the shorter rotation
+                data[CONF_LOOP_DIRECTION] = await self.hass.async_add_executor_job(
+                    get_pair_direction,
+                    self._pygtfs,
+                    data[CONF_ROUTE],
+                    data[CONF_ORIGIN].split(": ")[0],
+                    data[CONF_DESTINATION].split(": ")[0],
+                )
             _LOGGER.debug(f"UserInputs Destination: {data}")
             check_config = await self._check_config(data)
             if check_config == "stop_incorrect":
@@ -400,7 +414,7 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
             get_destination_stop_list,
             self._pygtfs,
             self._user_inputs[CONF_ROUTE],
-            self._user_inputs[CONF_DIRECTION],
+            None,
             self._user_inputs[CONF_ORIGIN].split(": ")[0],
         )
         if not destinations:
--- a/custom_components/gtfs2/config_flow.py
+++ b/custom_components/gtfs2/config_flow.py
@@ -56,6 +56,7 @@ from .gtfs_helper import (
     get_next_departure,
     get_route_list,
     get_stop_list,
+    get_destination_stop_list,
     get_datasources,
     remove_datasource,
     check_datasource_index,
@@ -75,6 +76,16 @@ TRANSLATION_DESCRIPTION_PLACEHOLDERS = {
     "docu_configuring_options": "https://github.com/vingerha/gtfs2/wiki/04:-Configuring-a-route's-options-(inc.-adding-real%E2%80%90time)",
     "model": "Example model",
 }
+
+
+def _stop_name_of(entry):
+    """The stop name inside a picker entry "stop_id: Name #2 (12)": the
+    number given to a repeated name and the sequence are display only."""
+    name = entry.split(": ", 1)[1].rsplit(" (", 1)[0]
+    if " #" in name and name.rsplit(" #", 1)[1].isdigit():
+        name = name.rsplit(" #", 1)[0]
+    return name
+
 
 @config_entries.HANDLERS.register(DOMAIN)
 class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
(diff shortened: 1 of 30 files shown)


vingerha:
Thanks and general note from me ... I am not sure if this is the curse or the benefit from AI. Its responses are often massive and thus difficult to grasp, making it complex to identify possible errors (you already corrected parts too in the other branch, things difficult to identify and imo unexepcted missed by AI). I asked AI to correct your tests for that branch 2 days ago (after my changes) and it just started to make sure that they did not fail, i.e. change script code or match input and output without even thinking about the idea behind it. 
You already know that I am more than a bit sceptical with AI as it always makes mistakes and/or interprets prompts slightly different then what I intended, plus it does not think of consequences unless you explicitly state all of this. This is making the good prompt super long and detailed, 
I perform quite a few analysis, builds, etc. with AI and where the original question may be 5 lines, in the end it is often 5 fold that number and needs to be redone in a new chat each time, since claude/chatgpt/gemini often forget earlier instructions or invent things on the fly
... anyhow me-ranting-about-AI but still see a huge gain, I will download it and test it a bit :)

Pulpyyyy:
AI is not so different from people, it just holds a lot more at once but keeps none of it over time. The maths exercises you did with your son yesterday surely had a statement that set the context and the goal and left no room for interpretation, not three lines thrown in a corner. A short prompt does not frame a project either. The AI will reach the goal one way or another, the work is keeping it from taking shortcuts and dropping the cases that already worked. That is also why the tests matter more to me than the description.

The other way round, explaining in three words what changes when several hundred lines of code move is not easier either. For a test: add a line from GVB 1, Zou 653 or TAO 22, pick any stop, and the second screen should only list what a trip really reaches from it. If anything looks wrong, the feed, the route and the two stops are enough for me to reproduce it.

vingerha:
Did a quick manual test 
- zou 926, look good
- zou 653, shows stops reversed to what I am used to but this is cosmetic 
- Palmbus B, looks good
- GVB Line 1, shows stops reversed to what I am used to but this is cosmetic 

The one thing is that I can not verify if all the stops work correctly...and I guess this is the complex bit
In the selection process only one start stop is shown and selected, the corresponding end station is found and this combination with unique stop_id is stored in the entity. Could it be (and I get more and more lost now) that there are other stops with the same name that go to the destination but since the ID is not stored in the entity, they will never be shown as an option to travel

Pulpyyyy:
Order for Zou 653 and GVB 1: it is one list for both directions, so it follows a trip and reads backwards for half the users. Only the display, the second screen follows the real trip. Could we ask the direction again, just after the departure selection (and only if the line has both directions), so the arrival list is sorted the way the user rides? Not tested, just an idea.

Other stops with the same name that you asked about: they are not lost. The entity stores one stop_id, but the query takes every stop of the same place with it (same parent station, or same name within about 150 m). For example Place Vauban on Palm Bus B has a pole on each side of the road, the entity stores one and the sensor gets the departures from both. A stop with the same name further away, like "Centre" in another village, is a different place and has its own entry in the list.

vingerha:
list order: I donot think that a good solution exists, see the palmbus B route description vs. outbound in opposite direction, which one is the truth and how can the correct(?) 'deafult' (?) direction be defined when using only the provider db. 
The only alternative I can think off is to have a tick box 'show reverse' that will show a next screen with the same stops reversed....but...well... no super enthusiastic

Other stops: great, I did not spend time on this.

So, from where I sit... - I think this could be wrapped up and brought live.
- it needs to be tested on update, i.e. when overwriting an latest release e.g. 0.5.9.10 (wrongly tagged, should have been 0.5.10), i.e. is it regression proof or does it need an entity migration or a 'breaking change' comment.
What is your view?

Pulpyyyy:
Asking direction would help in some cases : Departure in the middle of a long line: a user going towards the side shown second has to scroll past the first side. With the direction, their side would come first, or be the only one shown.

Pulpyyyy:
Seesm to me the code is no migration needed, older entries carry nothing the new code requires, and include_tomorrow is simply no longer read. One change deserves a line in the release notes: an entry now only shows its own line, where 0.5.9.10 showed every line serving the pair. Other than that, existing entries get more departures (the other side of the road), not fewer.

We need to cover both new and existing setup anyway

vingerha:
> Asking direction would help in some cases : Departure in the middle of a long line: a user going towards the side shown second has to scroll past the first side. With the direction, their side would come first, or be the only one shown.

But still, from the start - list you would not immediately know which direction that would be, i.e. for Zou 653 I now see the inbound one (most trips go Grasse>Antibes but it shows Antibes>Grasse)...but I only know this because I use it and only when I see the list I may (?) ask for a different order. So then a 'show-reverse'  button applies only then imo....or... two selection list with the same stops but in reverse order.. :) 

EDIT forget the two lists, then people select something in both...which to take :)

Pulpyyyy:
I think it can be solved from the provider db, without a reverse button, if the direction is only used to sort and split the lists, never to build them. Stops and trips stay built from the trips themselves, so a wrong label (GVB 1) still cannot hide anything.

Departure list: it is only there to say where you are, so a default order is enough. The Outbound order is what people are used to, so the list is turned round when most Outbound (direction 0) trips ride it backwards. On the test feeds that puts Zou 653 back to Grasse first and GVB 1 to Matterhorn first, and Palm Bus B stays as it is. Zou 926 would turn round, but it has one trip each way so there is not much to go on.

Then, only if the departure has stops on both sides, a short question "towards X / towards Y", X and Y being the last stop a trip really reaches on each side from that departure, not the end of the list (Zou 653 variants). The arrival list then only shows that half, nearest stops first. No question at the end of a line, on a loop (the fewest stops rule stays) or on rail.

The departure order is checked, the towards question is not built yet. If that works for you I add both to this PR.

Will be back later, stuffs needs to be done IRL

vingerha:
~~OK, so I think to read that your suggestion is to have a default direction_id 0 (not visible or selectable), that may do for most indeed, and we cannot serve all situations~~

For zou , currenlty dir 0 is Grasse>Antibes and for palmbus Cannes>Mouans, and gvb1 0 is Azart > Matterhorn (along gvb website too).... but again, I am not sure how you can identify which is which as these are only a few examples. Setting dir 0 by default should be fine I hope but also here, if not a good solution is available then I can live with any order ... or a button that says 'show reverse' to overcome any challenges that can not be foreseen

Pulpyyyy:
No direction_id stored in the entry, and the query does not change. There are two separate things.

The start list only needs a readable order, and that is the only place direction 0 plays a part: the list follows the way most direction 0 trips ride it. I do not need to know what 0 means on a network, it is counted per line. That gives Grasse first on 653 and Cannes first on B. On GVB 1 it gives Matterhorn first, because the feed labels both 455 Matterhorn > Azartplein trips and 324 Azartplein > Surinameplein trips as 0. Cosmetic, and no rule fixes a label like that.

The question is the actual idea, and it is only asked when it settles something: when the trips from the chosen departure really go both ways. Then "towards X / towards Y". At the end of a line there is only one way, so no question, same on a loop or on rail.

The question does not read direction_id either. The two sides come from where the trips really go from that stop, so it works the same when the labels are wrong, GVB 1 included. The answer picks the right way: the arrival list keeps only that side, nearest first.

So nobody has to find their way in a reversed list, and there is no need for a reverse button.

vingerha:
OK, well, you know your piece of code best. I was just thinking that at some point you have a list of stops which are derived without using direction_id, so how do you know which order they need to be, an additional piece of code that finds the most common 'start' for direction 0 ? No need to explain, I will await the update :) My claude reached its limit for this afternoon

Pulpyyyy:
Pushed the update. After the departure stop, the flow asks where the bus goes, but only when buses from that stop go to different termini. The arrival list then shows that side only, nearest first. At the end of a line there is no question.

The choice is the terminus of the trip, what the bus shows on its front, taken from the trips and not from direction_id, so GVB 1 works the same. On a loop both ways end at the same place, so the next stop is added: from Zénith on TAO 22 you get "Plissay" or "Jean Moulin", mid-loop "Vieux Poirier … Zénith" or "Bois Girault … Zénith", and at Zénith the answer also sets the rotation the entry keeps. Short workings are folded into the full line (GVB 1 Surinameplein, TAO 40 quai C). A line with real branches can show three choices, the same as the buses do.

Stops with several quays are handled as one stop: quays under the same parent station, or with the same name within about 150 m, are one entry (TAO tram A Hôpital de La Source, Zou 653 Pont de la Brague, the three Zénith poles). Departures come from every quay and each one says which quay it leaves from. When trips end on another quay of the terminus, that stays one choice (TAO 3 Belneuf, TAO 40 quai C).

The start list now begins at the end most direction 0 trips start from, as discussed.

vingerha:
Nice idea... will look at it later today

Pulpyyyy:
The arrival list now shows one branch at a time, busiest first: with branches A, B and C it used to read A1, B1, C1, A2, B2, C2, it now reads A1, A2, B1, B2, C1, C2. On Zou 653 from Grasse, RD du 24 Août no longer sits inside the Plascassier loop. Each bus keeps its own order, and the start list is unchanged.

vingerha:
Hmm, need to look at it (later). For destinations I would not know why selecting the end station requires multiple branches as I already know the desitnation which is fixed in the route entity...the entity just goes start/end so will show buses that go with and without the loop in plascassier (if the stop is after that)

... later, no need to answer until I see what you propose in real life :)

vingerha:
Some quick feedback, I did not look at the code: look very good to me
- zou 653 stop list nicely sorted (Grasse) and also works for stops that only offer a service in a single direction
- gvb 1 still starts with Matterhorn, ideally this should be Azartplein, cosmetic
- palmbus b, shows the list starting in Cannes (as known/current)

So, merge and bring this live?

Pulpyyyy:
Thanks for testing. One check: on Zou 926 the start list may now begin from the other end than in your first test. Is it still fine for you?

vingerha:
It is fine, I doubt that all cases can be covered...so merge & push?

vingerha:
and #211 can be closed then, correct?

There is a bit of documentation to be done before the release as I would like to refer to that...  I will do this update tomorrow I hope

Pulpyyyy:
Yep, #211 can be closed.

We'll never cover 100% of the cases perfectly, but if it's at least 95% human-readable, that's already a good result.

And that's exactly why mass automated tests will help us improve it, step by step.

PR #217, opened 2026-09-14
Improve route-based stop selecton and remove today/tomorrow mechanism

vingerha:
- route-based stop selection improved (see doc t.b.d)
- removed 'include tomorrow' so the entity will now show future departures, dot
- substantial code revision 
- substantial test improvement

Loads of the work by @Pulpyyyy


Pulpyyyy:
Don't forget #214

vingerha:
That went in automatically so it seems ... the code is in main. I must say that GH is sometimes a bit difficult to follow. One of my GH desktop connections crashed and this left 'supposedly' uncommited files on my local drive but all was committed, took me a while to verify that ...anyhow, 
All seems ok

vingerha:
@Pulpyyyy just did a bit on the documentation and seems only the route-based setup needed a bit of a rework, It is not pretty but seems covered now. For the time spent by us, I know it is a major improvement but the changes in the doc do not really demonstrate that. 
Tomorrow I will release it as a sub-release, just to make sure we have a few people that can test it too (no clue actually who downloads what / when and then ... if you just update, nothing changes)

Maybe some marketing on the main forum although I see a lot of city-specific transport apps being introduced too.

vingerha:
Maybe add a map to show where this is all working...

And, not sure how often I need to repeat this: thanks a lot for your work and 'drive' to move it forward

Pulpyyyy:
Merged, rebased, rebuilt (what a mess!)

And I found a few things that might need tweaking before release (only me ?!)
- On the start screen: the direction is no longer needed => remove “***”
- Missing translations
image

What if we chose more human-sounding names? No need (anymore) for technical details or numbers.
`ORLEANS:StopArea:THOPIT2: Hôpital de La Source (0)` => `Hôpital de La Source`
image

Pulpyyyy:
What i would use in fork 
image
image
image
image

Pulpyyyy:
What do you mean by a map? A world map in the README showing the networks where gtfs2 is known to work, or something in the integration itself?

vingerha:
The first. I am thinking to repost this integration but need a eyecatcher. Most people can't care less about the reach as long as it works for them which I guess 99% local only...so maybe not a good idea

PR #218, opened 2026-09-14
Open the datasource off the event loop

Pulpyyyy:
`get_gtfs` opens the sqlite file through pygtfs and, when the file is not there, downloads and unpacks the feed. Both coordinators and the two service helpers (`get_route_departures`, `get_trip_stops`) call it straight from a coroutine, so every refresh cycle does disk work on the event loop, and a datasource whose file has gone does a whole download there.

This hands the call to the executor, like the other blocking calls those functions already make. Four call sites, no other change. Your tests/ stay 17/17: the fake hass runs executor jobs inline, so they see the same call.


vingerha:
Thx, I will have another look at this, the current handling of extracting and covering for issues/erors is a bit of a mess
But, other stuff too, spent a lot of time here the past 1-2 weeks and will try to spend a bit more time on Music Assistant (want to have it multi-channel over hdmi)

vingerha:
Thx, I will add more guard steps in the code later.

